PR #2, opened 2021-12-17
Anx/rel

Anxhela21:
Appending related objects to search result.


jlpadilla:
Code was merged with PR #9

PR #17, opened 2022-02-07
Integration Test for Relationships

Anxhela21:
Continuation of unit tests for search-v2-api. Specifically this PR focuses on relationships.


pkg/resolver/search.go
```diff
@@ -144,17 +157,38 @@ func (s *SearchResult) resolveCount(query string, args []interface{}) (int, erro
...
+	defer rows.Close()
+	for rows.Next() {
+		err = rows.Scan(&uid)
+		if err != nil {
+			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), query)
+		}
+		s.uids = append(s.uids, &uid)
+	}
+
+	return s.uids, err
```
jlpadilla, line 176: There's no need to return `s.uids` from this function because you have saved it in the SearchResult object, so other functions can access it from there.

pkg/resolver/search.go
```diff
@@ -167,7 +201,7 @@ func (s *SearchResult) resolveItems(query string, args []interface{}) ([]map[str
 		currItem["cluster"] = cluster
 
 		items = append(items, currItem)
-		s.uids = append(s.uids, uid)
+		s.uids = append(s.uids, &uid)
```
jlpadilla, line 204: Here also need to declare `var uid string` inside the for loop.

pkg/resolver/search.go
```diff
@@ -186,29 +220,33 @@ func (s *SearchResult) getRelations() []SearchRelatedResult {
...
+		WHERE (e.sourceid = sg.destid OR e.destid = sg.sourceid)
+		AND r.uid <> all(sg.path)
+		AND level = 1
 		)
-	select distinct on (destid) data, destid, destkind from search_graph where level=1 or destid = ANY($2)`
+	SELECT distinct ON (destid) data, destid, destkind FROM search_graph WHERE level=1 OR destid = ANY($1)`)
 
-	relations, QueryError := s.pool.Query(context.Background(), recrusiveQuery, s.uids, s.uids) // how to deal with defaults.
+	fmt.Println("LENGTH OF QUERY:", len(relQuery))
+	fmt.Println("NUMBER OF UIDS IN RELATED:", len(s.uids))
```
jlpadilla, line 247: Use `klog.Info()`

pkg/resolver/testHelper_test.go
```diff
@@ -48,35 +49,75 @@ func (r *Row) Scan(dest ...interface{}) error {
...
+		mockData := make([]map[string]interface{}, len(items))
+		for i, item := range items {
+			uid := item.(map[string]interface{})["uid"]
+			mockData[i] = map[string]interface{}{
+				"uid":     uid,
+				"cluster": strings.Split(uid.(string), "/")[0],
+				"data":    item.(map[string]interface{})["properties"],
+			}
+		}
+		columnHeaders := []string{"uid string", "cluster string", "data *map[string]interface{}"}
```
jlpadilla, line 71: Remove the types from here because Scan() is inferring it from the arg type.  I think that's better than declaring it here.

pkg/resolver/testHelper_test.go
```diff
@@ -92,13 +133,60 @@ func (r *MockRows) Next() bool {
...
+		}
+	}
 	return nil
 }
 
 func (r *MockRows) Values() ([]interface{}, error) { return nil, nil }
 
 func (r *MockRows) RawValues() [][]byte { return nil }
+
+func mergeMaps(maps ...[]map[string]interface{}) []map[string]interface{} {
```
jlpadilla, line 184: Is this used anywhere?

pkg/resolver/testHelper_test.go
```diff
@@ -48,35 +49,75 @@ func (r *Row) Scan(dest ...interface{}) error {
...
+		var data map[string]interface{}
+		if err := json.Unmarshal(bytes, &data); err != nil {
+			panic(err)
+		}
+		items := data["addResources"].([]interface{})
+		edges := data["addEdges"].([]interface{})
+
+		mockResources := make([]map[string]interface{}, len(items))
+		for i, item := range items {
+			uid := item.(map[string]interface{})["uid"]
```
jlpadilla, line 90:
```suggestion
			uid := item.(map[string]interface{})["uid"].(string)
```

pkg/resolver/testHelper_test.go
```diff
@@ -92,13 +133,60 @@ func (r *MockRows) Next() bool {
...
+
 func (r *MockRows) Scan(dest ...interface{}) error {
-	*dest[0].(*string) = r.mockData[r.index-1]["uid"].(string)
-	*dest[1].(*string) = r.mockData[r.index-1]["cluster"].(string)
-	*dest[2].(*map[string]interface{}) = r.mockData[r.index-1]["data"].(map[string]interface{})
+	for _, col := range r.columnHeaders {
+		for _, value := range dest {
+			switch v := value.(type) {
+			default:
+				fmt.Printf("unexpected type %T", v)
```
jlpadilla, line 158: The default case might need to be at the end. (I could be mixing it with JS).

pkg/resolver/mocks/mock-rel-1.json
```diff
@@ -0,0 +1,18 @@
+{"addRecords":[
```
jlpadilla, line 1: This can be named just "records" (or "rows" if you want to match the postgres language).

pkg/resolver/mocks/mock-rel-1.json
```diff
@@ -0,0 +1,18 @@
...
+    {
+        "uid":"local-cluster/e12c2ddd-4ac5-499d-b0e0-20242f508afd",
```
jlpadilla, line 3: You can also simplify the UIDs. It could be something like this.
```
"local-cluster/00000000-0000-0000-0000-000000001111"
```

pkg/resolver/mocks/mock-rel.json
```diff
@@ -0,0 +1,41 @@
+
+{
```
jlpadilla, line 2: This look different from `mock-rel-1.json`.  Do you intend to use the same format as `mock-rel-1.json` here?

pkg/resolver/search.go
```diff
@@ -105,6 +109,7 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool) (string
 
 		if len(filter.Values) > 1 {
 			for _, val := range filter.Values {
+				klog.Info("filter VALUE is:", val)
 				klog.Infof("Filter value: %s", *val)
```
jlpadilla, line 113: Duplicate log message.

PR #18, opened 2022-02-08
Parameterize all user inputs for general search

SherinV:
Parameterize inputs
Add test for multiple filters
For https://github.com/stolostron/backlog/issues/18041 and https://github.com/stolostron/backlog/issues/19519


pkg/resolver/search.go
```diff
@@ -42,6 +44,7 @@ func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, e
 }
 
 func (s *SearchResult) Count() int {
+	klog.Info("Resolving SearchResult:Count()")
```
jlpadilla, line 47: Let's make this log v=2 or higher.
```
klog.V(2).Info("Resolving SearchResult:Count()")
```

pkg/resolver/search.go
```diff
@@ -77,62 +80,36 @@ func (s *SearchResult) Related() []SearchRelatedResult {
...
+	var selectDs *goqu.SelectDataset
+	var whereDs []exp.Expression
+
+	// Example query: SELECT uid, cluster, data FROM search.resources  WHERE lower(data->> 'kind') IN
+	// (lower('Pod')) AND lower(data->> 'cluster') IN (lower('local-cluster')) LIMIT 10000
+	//FROM CLAUSE
+	schemaTable := goqu.S("search").Table("resources")
+	ds := goqu.From(schemaTable)
+	//SELECT CLAUSE
+	selectDs = ds.Select("uid", "cluster", "data")
```
jlpadilla, line 94: Move this to else to avoid reassigning.
```
if count {
	selectDs = ds.Select(goqu.COUNT("uid"))
} else {
	selectDs = ds.Select("uid", "cluster", "data")
}
```

pkg/resolver/search.go
```diff
@@ -77,62 +80,36 @@ func (s *SearchResult) Related() []SearchRelatedResult {
...
-		query = selectClause + strings.TrimRight(whereClause, trimAND)
+		limit = config.DEFAULT_QUERY_LIMIT
 	}
-	klog.Infof("query: %s\nargs: %+v", query, args)
-
-	return query, args
+	//Get the query
+	sql, params, err := selectDs.Where(whereDs...).Limit(uint(limit)).ToSQL()
+	if err != nil {
+		klog.Infof("Error building SearchComplete query: %s", err.Error())
```
jlpadilla, line 109: Change to `klog.Error` or `klog.Warn`

pkg/resolver/search.go
```diff
@@ -283,6 +260,15 @@ func formatLabels(labels map[string]interface{}) string {
 	return strings.Join(labelStrings, ",")
 }
 
+func formatArray(labels []interface{}) string {
```
jlpadilla, line 263: Add comment
```
// Encode labels array into a single string with the format.
//  key=value, key2=value2, ...
```

pkg/resolver/search.go
```diff
@@ -283,6 +260,15 @@ func formatLabels(labels map[string]interface{}) string {
...
+	keys := make([]string, 0)
+	for _, k := range labels {
+		keys = append(keys, fmt.Sprintf("%s", k))
+	}
```
jlpadilla, line 267: We know the size of the array, so we can allocate it. *We won't ever notice the performance difference here, it's just for best practice.
```go
keys := make([]string, len(labels))
	for i, k := range labels {
		keys[i] = fmt.Sprintf("%s", k) //<-- I don't think this will work for non-strings.
	}
```

jlpadilla: Wait, why the diff isn't showing what's currently on main?
https://github.com/stolostron/search-v2-api/blob/main/pkg/resolver/search.go#L270-L284

pkg/resolver/search.go
```diff
@@ -295,10 +281,57 @@ func formatDataMap(data map[string]interface{}) map[string]interface{} {
...
+			item[key] = formatArray(v)
 		default:
 			klog.Warningf("Error formatting property with key: %+v  type: %+v\n", key, reflect.TypeOf(v))
 			continue
 		}
 	}
 	return item
 }
+
+func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
```
jlpadilla, line 294: lowercase method name to keep private `whereClauseFilter`

pkg/schema/searchComplete.go
```diff
@@ -2,61 +2,104 @@ package schema
...
+		}
+		//Get the query
+		sql, params, err := selectDs.Where(whereDs...).Limit(uint(limit)).ToSQL()
+		if err != nil {
+			klog.Infof("Error building SearchComplete query: %s", err.Error())
 		}
-		klog.Info("SearchComplete Query: ", query)
-		return query
+		klog.Infof("SearchComplete: %s\nargs: %s", sql, params)
+		return sql, params
```
jlpadilla, line 80: Instead of returning query and params, you could save in `s`.  It reduces the number of parameters that you need to pass around.

pkg/resolver/search.go
SherinV: I am using that function in searchComplete file too. That is why it is public.

pkg/schema/searchComplete.go
```diff
@@ -2,61 +2,104 @@ package schema
...
 	//TODO: Handle error
-	rows, _ := pool.Query(context.Background(), query)
+	rows, err := s.pool.Query(context.Background(), query, args...)
+	if err != nil {
+		klog.Error("Error fetching results from db ", err)
+	}
 	defer rows.Close()
 	var srchCompleteOut []*string
-	prop := ""
+	var prop string
```
jlpadilla, line 95: - define the variable prop inside the for loop.
- Then you can delete this line. 
  ```
  tmpProp := prop
  ```

pkg/schema/testHelper_test.go
```diff
@@ -0,0 +1,100 @@
+// Copyright Contributors to the Open Cluster Management project
+package schema
```
jlpadilla, line 2: Let's try to avoid duplicating this file.
- Could we reference these from the resolver package?
- Could we combine the schema and resolver packages into a single package?

SherinV: There are a few differences between the two. is it possible to overload the Scan Method and Row structs?

pkg/resolver/search.go
```diff
@@ -295,10 +281,57 @@ func formatDataMap(data map[string]interface{}) map[string]interface{} {
...
+	var colDs exp.IdentifierExpression
```
jlpadilla, line 295: Would it possible to combine the `goqu.Ex{}` syntax with `jsonb` ?  It could make this function much simpler.
Here's the documentation for the syntax I'm thinking: http://doug-martin.github.io/goqu/docs/selecting.html#where

The query will look something like this (without considering jsonb)
```go
.Where(goqu.Ex{
  "kind": string[]{"pod", "deployment"},
  "name": string[]{"name-a", "name-b"},
})
```

SherinV: I haven't found how to combine json and `goqu.Ex`. But, simplified the whereClause function with `goqu.L` for json columns.

pkg/schema/searchComplete.go
SherinV: Done

pkg/schema/searchComplete.go
SherinV: Done

pkg/resolver/search.go
SherinV: Done

pkg/resolver/search.go
SherinV: Done

pkg/resolver/search.go
```diff
@@ -77,77 +82,55 @@ func (s *SearchResult) Related() []SearchRelatedResult {
...
+	//WHERE CLAUSE
+	if s.input != nil && len(s.input.Filters) > 0 {
+		whereDs = WhereClauseFilter(s.input)
 	}
-	if limitStr != "" {
-		limitClause = " LIMIT " + limitStr
-		query = selectClause + strings.TrimRight(whereClause, trimAND) + limitClause
-
+	//LIMIT CLAUSE
+	if s.input != nil && s.input.Limit != nil && *s.input.Limit != 0 {
```
jlpadilla, line 106: I **think** that the API will accept negative values, to be safe lets check limit is > 0.
```go
if s.input != nil && s.input.Limit != nil && *s.input.Limit > 0 {
```

jlpadilla: Also if `count == true` then a LIMIT doesn't apply.

pkg/resolver/search.go
```diff
@@ -295,10 +305,30 @@ func formatDataMap(data map[string]interface{}) map[string]interface{} {
...
+		if len(filter.Values) > 0 {
+			for _, val := range filter.Values {
+				values = append(values, *val)
+			}
+		}
+		if filter.Property == "cluster" {
+			whereDs = append(whereDs, goqu.C(filter.Property).In(values).Expression())
+		} else {
+			whereDs = append(whereDs, goqu.L(`"data"->>?`, filter.Property).In(values).Expression())
+		}
```
jlpadilla, line 331: Couple changes:
- Check if a filter has no values.
- Allocate array instead of append()

```suggestion
		if len(filter.Values) > 0 {
			values := make([]string, len(filter.Values))
			for i, val := range filter.Values {
				values[i] = *val
			}
		
			if filter.Property == "cluster" {
				whereDs = append(whereDs, goqu.C(filter.Property).In(values).Expression())
			} else {
				whereDs = append(whereDs, goqu.L(`"data"->>?`, filter.Property).In(values).Expression())
			}
		} else {
			klog.Warnf("Ignoring filter [%s] because it has no values", filter.Property)
		}
```

jlpadilla: The graphQL schema should also enforce that values isn't empty.  We can add `!` here.

https://github.com/stolostron/search-v2-api/blob/main/graph/schema.graphqls#L27

pkg/resolver/search.go
```diff
@@ -283,6 +266,33 @@ func formatLabels(labels map[string]interface{}) string {
...
+	keys := make([]string, len(itemlist))
+	for i, k := range itemlist {
+		keys[i] = formatItem(k)
+	}
+	sort.Strings(keys)
+	return strings.Join(keys, ",")
+}
+
+// Convert interface to string format
+func formatItem(data interface{}) string {
```
jlpadilla, line 281: Suggestion: Would `convertToString()` be a better name?

pkg/schema/searchComplete.go
```diff
@@ -2,61 +2,104 @@ package schema
...
+
+	//FROM CLAUSE
+	schemaTable := goqu.S("search").Table("resources")
+	ds := goqu.From(schemaTable)
+	//WHERE CLAUSE
+	if s.input != nil && len(s.input.Filters) > 0 {
+		whereDs = res.WhereClauseFilter(s.input)
+	}
+	//SELECT CLAUSE
+	if s.property != "" {
```
jlpadilla, line 59: We could move this `if` to the start of this function.  If we don't have this value we can't do anything.

pkg/schema/searchComplete_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// Mock the database queries.
+	mockRows := newMockRows("../resolver/mocks/mock.json")
+	// Mock the database query
+	mockPool.EXPECT().Query(gomock.Any(),
+		gomock.Eq(`SELECT DISTINCT "data"->>'kind' FROM "search"."resources" WHERE ("data"->>'kind' IS NOT NULL) ORDER BY "data"->>'kind' ASC LIMIT 10000`),
+		gomock.Eq([]interface{}{})).Return(mockRows, nil)
+
+	// Execute function
+	result, err := resolver.autoComplete(context.TODO())
+	fmt.Println("err from autoComplete: ", err)
```
jlpadilla, line 31: Suggestion: It's better to use `t.Log()`

pkg/schema/searchComplete_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func string_array_equal(result, expected []*string) bool { //, expected []interface{}) bool {
+	for i, exp := range expected {
+		if *result[i] != *exp {
+			return false
+		}
+	}
+	return true
+}
```
jlpadilla, line 70: This is a nice function to reuse across tests. I'm just adding a similar one to the indexer [in this PR](https://github.com/stolostron/search-indexer/pull/31/files#diff-ee17c67bdc69ad2fdca5d64f2a447c755f5efbf88df225b38aa29826c2bd8dedR19-R24)

I want to suggest a few changes to keep consistency.
```suggestion
func AssertStringArrayEqual(t *testing.T, result, expected []*string, message string) {
	for i, exp := range expected {
		if *result[i] != *exp {
			t.Errorf("%s expected [%v] got [%v]", message, expected, result)
			return
		}
	}
}
```

pkg/schema/searchComplete.go
SherinV: Done

pkg/resolver/search.go
SherinV: Done

pkg/resolver/search.go
SherinV: Done

pkg/resolver/search.go
SherinV: Done

pkg/resolver/search.go
```diff
@@ -77,81 +82,63 @@ func (s *SearchResult) Related() []SearchRelatedResult {
...
-				// "SELECT id FROM search.resources WHERE status = any($1)"
-				//SELECT id FROM search.resources WHERE status = ANY('{"Running", "Error"}');
-			}
-			whereClause = whereClause + "=any($" + strconv.Itoa(i+1) + ") AND "
-			args = append(args, pq.Array(values))
-		} else if len(filter.Values) == 1 {
-			whereClause = whereClause + "=$" + strconv.Itoa(i+1) + " AND "
-			val := filter.Values[0]
-			args = append(args, strings.ToLower(*val))
+			fmt.Println("final ELSE")
```
jlpadilla, line 115: Please use `klog` or remove.

SherinV: Missed it. Done.

jlpadilla (approved):
/lgtm

PR #19, opened 2022-02-22
Separate resolver function for fetching uids.

Anxhela21:


pkg/resolver/search.go
```diff
@@ -133,6 +143,26 @@ func (s *SearchResult) resolveCount() (int, error) {
 	return count, err
 }
 
+func (s *SearchResult) resolveUids() ([]*string, error) {
+	rows, err := s.pool.Query(context.Background(), s.query, s.params...)
+
+	if err != nil {
+		klog.Errorf("Error resolving query [%s] with args [%+v]. Error: [%+v]", s.query, s.params, err)
+	}
+	var uid string
```
jlpadilla, line 152: Declare this variable inside the for loop below.

pkg/resolver/search.go
```diff
@@ -133,6 +143,26 @@ func (s *SearchResult) resolveCount() (int, error) {
 	return count, err
 }
 
+func (s *SearchResult) resolveUids() ([]*string, error) {
+	rows, err := s.pool.Query(context.Background(), s.query, s.params...)
+
+	if err != nil {
+		klog.Errorf("Error resolving query [%s] with args [%+v]. Error: [%+v]", s.query, s.params, err)
```
jlpadilla, line 150: I wonder if we are able to process anything else when we have an error here or if we should stop and return the error.

pkg/resolver/search.go
```diff
@@ -70,18 +70,28 @@ func (s *SearchResult) Items() []map[string]interface{} {
...
 	}
-	s.wg.Wait() // WORKAROUND wait to complete execution of Items()
+
+	s.wg.Wait()
 
 	r := s.getRelations()
 	return r
 }
 
+func (s *SearchResult) Uids() []*string {
```
jlpadilla, line 83: No need to return the `[]*string` because we are saving the result in the SearchResult object.

pkg/resolver/search.go
Anxhela21: Do you mean for this specific resolver or for all of the resolvers?

jlpadilla: It applies in other places too.  I don't think `rows` will have anything if there's an error, maybe we could get partial results?

pkg/resolver/search.go
```diff
@@ -133,6 +142,26 @@ func (s *SearchResult) resolveCount() (int, error) {
...
+	for rows.Next() {
+		var uid string
+		err = rows.Scan(&uid)
+		if err != nil {
+			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), s.query)
+		}
+		s.uids = append(s.uids, &uid)
+	}
+
+	return s.uids, err
```
SherinV, line 162: These returned uids are not being used in line 86. The error is already logged here in line 157.

> _, e := s.resolveUids()

We don't need to return those, in that case.

pkg/resolver/search.go
```diff
@@ -70,18 +70,27 @@ func (s *SearchResult) Items() []map[string]interface{} {
...
-	s.wg.Wait() // WORKAROUND wait to complete execution of Items()
+
+	s.wg.Wait()
 
 	r := s.getRelations()
 	return r
 }
 
+func (s *SearchResult) Uids() {
+	klog.Info("Resolving SearchResult:Uids()")
```
jlpadilla, line 84: Let's only log this when running in verbose mode.
```
klog.V(3).Info("Resolving SearchResult:Uids()")
```

pkg/resolver/search.go
Anxhela21: good catch!

jlpadilla (approved):
/lgtm
/hold

jlpadilla (approved):
/lgtm

While implementing this change, we realized that a better alternative is to update the recursive query to use the search input instead of uids.  We agreed to merge this improvement and work on the better solution after.

PR #21, opened 2022-02-28
Implementing Keyword Search

Anxhela21:
Related Issue: [https://github.com/stolostron/backlog/issues/19691](https://github.com/stolostron/backlog/issues/19691)


pkg/resolver/search.go
```diff
@@ -45,14 +45,16 @@ func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, e
 	return srchResult, nil
 }
 
+func (s *SearchResult) KeywordSearch() { //need this function and resolver for tests.
+	klog.V(2).Info("Resolving SearchResult:Keywords()")
+	s.buildSearchQuery(context.Background(), false, false)
```
jlpadilla, line 50: change to `context.TODO()`

pkg/resolver/search.go
```diff
@@ -45,14 +45,16 @@ func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, e
 	return srchResult, nil
 }
 
+func (s *SearchResult) KeywordSearch() { //need this function and resolver for tests.
```
jlpadilla, line 48: Why do we need a different function?  A search query could include both keywords and filters so we need to process from the same resolver.

pkg/resolver/search.go
```diff
@@ -96,20 +98,46 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
...
+	if s.input.Keywords != nil {
+		schemaTable := goqu.S("search").Table("resources")
+		jsb := goqu.L("jsonb_each_text(?)", goqu.C("data"))
+		ds := goqu.From(schemaTable, jsb)
+		if count {
+			selectDs = ds.Select(goqu.COUNT("uid"))
+		} else if uid {
+			selectDs = ds.Select("uid")
+		} else {
+			selectDs = ds.Select("uid", "cluster", "data", "key", "value")
```
jlpadilla, line 110: We don't need to return `key` and `value`.  We use those only for matching, those will be included in `data`.

pkg/resolver/search.go
```diff
@@ -130,13 +158,19 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
...
 	rows := s.pool.QueryRow(context.Background(), s.query, s.params...)
 
 	var count int
 	err := rows.Scan(&count)
+	if err != nil {
+		klog.Errorf("Error %s resolving count for query:%s", err.Error(), s.query)
+	}
+
+	s.count = count
+	count = s.count
```
jlpadilla, line 171: I don't think we use `s.count` so we don't need this and we can remove `count` from the SearchResult struct.

pkg/resolver/search.go
```diff
@@ -160,7 +194,7 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
 	if err != nil {
 		klog.Errorf("Error resolving query [%s] with args [%+v]. Error: [%+v]", s.query, s.params, err)
 	}
-	defer rows.Close()
+	// defer rows.Close()
```
jlpadilla, line 197: We need to call row.Close() to avoid leaking connections in the pool.

pkg/resolver/search.go
```diff
@@ -169,7 +203,13 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
 
 	for rows.Next() {
 		var uid string
-		err = rows.Scan(&uid, &cluster, &data)
+		if s.input.Keywords != nil {
+			var key *string
+			var value *interface{}
+			err = rows.Scan(&uid, &cluster, &data, &key, &value)
+		} else {
```
jlpadilla, line 210: We don't need to return key and value.

pkg/resolver/search.go
```diff
@@ -344,21 +385,43 @@ func formatDataMap(data map[string]interface{}) map[string]interface{} {
...
+				}
 
-			if filter.Property == "cluster" {
-				whereDs = append(whereDs, goqu.C(filter.Property).In(values).Expression())
-			} else {
-				whereDs = append(whereDs, goqu.L(`"data"->>?`, filter.Property).In(values).Expression())
 			}
 		} else {
-			klog.Warningf("Ignoring filter [%s] because it has no values", filter.Property)
+			klog.Warningf("Ignoring filter [%s] because it has no values", input.Keywords)
```
jlpadilla, line 404: Small update to this message.
```suggestion
			klog.Warningf("Ignoring keyword filter [%s] because it has no values", input.Keywords)
```

pkg/resolver/search.go
Anxhela21: This was supposed to be removed. Thanks for catching that.

pkg/resolver/search.go
Anxhela21: Okay wasn't sure about that, thanks

pkg/resolver/search.go
Anxhela21: forgot to uncomment, thanks.

pkg/resolver/search.go
```diff
@@ -390,18 +416,38 @@ func pointerToStringArray(pointerArray []*string) []string {
...
 		} else {
-			klog.Warningf("Ignoring filter [%s] because it has no values", filter.Property)
+			klog.Warningf("Ignoring keyword filter [%s] because it has no values", input.Keywords)
 		}
 	}
+	if input.Filters != nil {
+		for _, filter := range input.Filters {
+			if len(filter.Values) > 0 {
+				values := pointerToStringArray(filter.Values)
+				for i, val := range filter.Values {
```
SherinV, line 437: This is what the function does above

pkg/resolver/search.go
```diff
@@ -101,20 +96,46 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
...
+		} else {
+			selectDs = ds.Select("uid", "cluster", "data")
+		}
 	}
 	//WHERE CLAUSE
 	if s.input != nil && len(s.input.Filters) > 0 {
 		whereDs = WhereClauseFilter(s.input)
 	}
+	if s.input.Keywords != nil { //if keyword is input then
+		if s.input != nil && len(s.input.Keywords) > 0 { //if the input is not nil and the len of keywords is greater than 0
```
SherinV, line 128: Combine lines 127 and 128

pkg/resolver/search.go
```diff
@@ -101,20 +96,46 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
...
+			whereDs = WhereClauseFilter(s.input) //the where clause will be
+			sql, params, err := selectDs.Where(whereDs...).Limit(uint(limit)).ToSQL()
```
SherinV, line 130: The query is built later in the function. Remove to avoid duplication and error

pkg/resolver/search.go
```diff
@@ -175,6 +185,7 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
 	for rows.Next() {
 		var uid string
 		err = rows.Scan(&uid, &cluster, &data)
+		// }
```
jlpadilla, line 188: delete this line

pkg/resolver/searchComplete.go
```diff
@@ -48,6 +48,18 @@ func (s *SearchCompleteResult) searchCompleteQuery(ctx context.Context) {
 	var selectDs *goqu.SelectDataset
 
 	//FROM CLAUSE
+	// if s.input.Keywords != nil && len(s.input.Keywords) > 0 {
```
jlpadilla, line 51: Why is this block commented? Remove if not needed or add explanation in the code if it's something that we'll need later.

pkg/resolver/searchComplete.go
```diff
@@ -105,3 +117,54 @@ func (s *SearchCompleteResult) searchCompleteResults() ([]*string, error) {
 	}
 	return srchCompleteOut, nil
 }
+
+// func (s *SearchCompleteResult) searchCompleteQuery(ctx context.Context) {
```
jlpadilla, line 121: Same here, why is this function commented?

SherinV (approved):

PR #22, opened 2022-03-01
Rewrite relations query with goqu

SherinV:

For https://github.com/stolostron/backlog/issues/20282
Rewrite relations query with goqu template
~Add relatedKinds filter~

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -7,10 +7,10 @@ package config
 
 import (
 	"encoding/json"
+	"errors"
 	"net/url"
 	"os"
 	"strconv"
-	"strings"
 
 	klog "k8s.io/klog/v2"
 )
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
```diff
@@ -7,10 +7,10 @@ package config
```
jlpadilla, line 7: Thanks for making this consistent with indexer!

pkg/resolver/search.go
```diff
@@ -186,39 +191,50 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
+	//The level can be parameterized later, if needed, for applications
+	whereDs = append(whereDs, goqu.C("level").Eq(1)) // Add filter to select only level 1 relationships
 
 	//defining variables
 	items := []map[string]interface{}{}
 	var kindSlice []string
 	var kindList []string
 	var countList []int
 
-	relQuery := strings.TrimSpace(`WITH RECURSIVE
```
jlpadilla, line 201: Should we keep a comment with the original query for documentation?

pkg/resolver/search.go
```diff
@@ -186,39 +191,50 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
+					InnerJoin(goqu.T("search_graph").As("sg"),
+						goqu.On(goqu.ExOr{"sg.destid": goqu.I("e.sourceid"), "sg.sourceid": goqu.I("e.destid")})).
+					Select(selectNext...).
+					Where(goqu.Ex{"sg.level": goqu.L("1"),
+						"r.uid": goqu.Op{"neq": goqu.All("{sg.path}")}}))).
+		Select("data", "destid", "destkind").Distinct("destid").
+		Where(whereDs...).ToSQL()
+
+	if err != nil {
+		klog.Error("Error creating relation query", err)
```
jlpadilla, line 235: I wonder if we are able to continue after this error.

pkg/resolver/search.go
```diff
@@ -186,39 +191,50 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
+	}
+	relations, relQueryError := s.pool.Query(context.Background(), sql, params...) // how to deal with defaults.
```
jlpadilla, line 237: context.TODO()

pkg/resolver/search.go
SherinV: Done

pkg/resolver/search.go
jlpadilla: Would the client receive an error? GraphQL can send partial results and errors.
Also the error we send to the client shouldn't describe any internals like the SQL query.

**We can move this discussion out of this PR, we should review the error path for the entire project.**

SherinV: Yes, we will revisit the error path for the entire api along with the context.

jlpadilla (approved):
/lgtm
/hold

PR #25, opened 2022-03-22
Fix duplicate results error

SherinV:

The issue was in the `buildSearchQuery`. Adding `jsonb_each_text(data)` in the from clause when there were no keywords caused duplicates.


Anxhela21 (approved):
/lgtm

PR #28, opened 2022-04-12
Authenticate token

Anxhela21:

Related User Story: [17982](https://github.com/stolostron/backlog/issues/17982)

Part 1 of the authenticate/authorize requests user story.


main.go
```diff
@@ -27,9 +32,20 @@ func main() {
 		klog.Fatal(configError)
 	}
 
+	//Check and verify token:
+	config.KubeClient()
+
 	database.GetConnection()
+
+	router := chi.NewRouter()
```
jlpadilla, line 40: The indexer is using the [gorilla/mux router](https://github.com/gorilla/mux). I want to be consistent across our projects.

I haven't evaluated these libraries. It's ok to change if `chi` is better, but we should update our other projects for consistency.

main.go
```diff
@@ -27,9 +32,20 @@ func main() {
...
+	router.Use(config.Middleware())
+	router.Handle("/", playground.Handler("search", "/query"))
+	log.Fatal(http.ListenAndServe("localhost:4010", router))
```
jlpadilla, line 43: Should move this into the logic below.

pkg/config/middleware.go
```diff
@@ -0,0 +1,136 @@
+package config
```
jlpadilla, line 1: Let's separate this logic in 2 files.
- config/kubeClient.go
- rbac/authMiddleware.go

Note the different package names. I want to start a new package to keep everything related to rbac.

pkg/config/middleware.go
```diff
@@ -0,0 +1,136 @@
...
+				return
+			}
+			fmt.Fprintf(w, "Authentication successful!")
+
+		})
+	}
+}
+
+func verifyToken(clientId string) (bool, error) {
+	ctx := context.TODO()
```
jlpadilla, line 115: Use the context from the request object.  https://pkg.go.dev/net/http#Request.Context

main.go
Anxhela21: Agree we should have consistency. I only chose chi because it seems to be the most updated of the two projects and it looks like mux is looking for [new maintainers](https://github.com/gorilla/mux/issues/659).

pkg/config/middleware.go
```diff
@@ -0,0 +1,136 @@
...
+		klog.V(2).Info("Creating k8s client using InClusterClientConfig")
+		clientConfig, clientConfigError = rest.InClusterConfig()
+	}
+
+	if clientConfigError != nil {
+		klog.Fatal("Error getting Kube Config: ", clientConfigError)
+	}
+
+	token := clientConfig.BearerToken
+	fmt.Printf("The bearer token is: %s\n", token)
```
jlpadilla, line 69: No need to return `token` because you return clientConfig, which contains the BearerToken.

**Note this is our Search API's token.**

pkg/config/middleware.go
```diff
@@ -0,0 +1,136 @@
...
+	return clientConfig, clientConfigError, token
+}
+
+//verifies token (userid) with the TokenReview:
+func Middleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+			token := KubeClient()
+			fmt.Printf("The token is: %s", token)
```
jlpadilla, line 80: This is our **Search API's Token**.  However, here we are in the context of a user request.  The user's token should be in the Authorization Header or cookie.
```go
r.Header.Get("Authorization")
```

jlpadilla: Here's the doc to read the cookies (I believe that requests from the UI have it in a cookie instead of the header.
- https://pkg.go.dev/net/http#Request.Cookie

pkg/config/middleware.go
```diff
@@ -0,0 +1,136 @@
...
+			authenticated, err := verifyToken(clientId)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			if !authenticated {
+				http.Error(w, "Invalid token", http.StatusForbidden)
+				return
+			}
+			fmt.Fprintf(w, "Authentication successful!")
```
jlpadilla, line 108: This line sends a response to the client, which completes the request. From here, we should only send a response to the client if there's an error.
If authentication is successful, we call `next()` to continue processing the request.

pkg/config/middleware.go
Anxhela21: Okay, wasn't sure here because I didn't see authorization header in request but did see cookie, so we can use that.

jlpadilla: We'll need to check both to cover the different scenarios.

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+//verifies token (userid) with the TokenReview:
+func Middleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+			// Read the value of the client identifier from the request header
+			fmt.Fprintf(w, "%s %s %s \n", r.Method, r.URL, r.Proto)
+
+			//if there is cookie in header available use that else use the authorization:
+			var tokenId string
```
jlpadilla, line 41: Let's call this `userToken` to be more explicit and avoid confusion later.

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+			if len(r.Cookies()) != 0 {
+				for _, cookie := range r.Cookies() {
+					if cookie.Name == "acm-access-token-cookie" {
+						log.Println("Cookie: ", cookie.Value)
```
jlpadilla, line 45: - We can't log the token per secure engineer requirements, but we should log that we found the token in the cookie.
- Use klog for consistency.
- Use a higher verbosity level, to avoid logging too much.
```suggestion
						klog.V(5).Info("Got user token from cookie.")
```

Anxhela21: Okay. Do we want to log the API TokenReview result?
ex:
```
{
        "authenticated": true,
        "user": {
                "username": "kube:admin",
                "groups": [
                        "system:cluster-admins",
                        "system:authenticated"
                ],
                "extra": {
                        "scopes.authorization.openshift.io": [
                                "user:full"
                        ]
                }
        },
        "audiences": [
                "https://kubernetes.default.svc"
        ]
}
```

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+						tokenId = cookie.Value
+					}
+				}
+			} else {
+				if len(r.Header.Get("Authorization")) != 0 {
+					log.Println(w, "Authorization is: %s", r.Header.Get("Authorization"))
+					tokenId = r.Header.Get("Authorization")
+				}
```
jlpadilla, line 53: - combine `else if`.
- same as previous comment about logging.
- respond with error if request doesn't contain the token.
```suggestion
			} else if len(r.Header.Get("Authorization")) != 0 {
				klog.V(5).Info("Got user token from Authorization header.")
				tokenId = r.Header.Get("Authorization")
			} else {
				http.Error(w, "Could not find a valid token.", http.StatusUnauthorized)
				return
			}
```

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+			//Iterate over all header fields
+			for k, v := range r.Header {
+				fmt.Fprintf(w, "header key %q, value %q\n", k, v)
+			}
```
jlpadilla, line 58: Looks like this was here for debug. Delete if not needed.

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+			//Retrieving and verifying the token:
+			// clientId := r.Header.Get("Userid")
+			if len(tokenId) == 0 {
+				http.Error(w, "Could not find Userid", http.StatusUnauthorized)
+				return
+			}
+			authenticated, err := verifyToken(tokenId, r)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
```
jlpadilla, line 67: - Log a warning in the server, for us to debug the problem. This can contain the actual error.
- Return a generic error message to the client per secure engineering best practices. This is because the `err.Error()` message could reveal too much details about the backend implementation.

```suggestion
				klog.Warning("Error authenticating the user token.", err.Error())
				http.Error(w, "Unexpected error while authenticating the request", http.StatusInternalServerError)
```

pkg/rbac/middleware.go
jlpadilla: or `clientToken`

pkg/rbac/middleware.go
jlpadilla: Actually, maybe it's better to log the error inside `verifyToken()`

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+func verifyToken(clientId string, r *http.Request) (bool, error) {
+	ctx := context.Context(r.Context())
+	tr := authv1.TokenReview{
+		Spec: authv1.TokenReviewSpec{
+			Token: clientId,
+		},
+	}
+	result, err := kClientset.AuthenticationV1().TokenReviews().Create(ctx, &tr, metav1.CreateOptions{})
+	if err != nil {
+		return false, err
```
jlpadilla, line 89: Log the error here so we can use it to debug the problem.
```golang
klog.Warning("Error authenticating the user token.", err.Error())
```

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	log.Printf("%v\n", prettyPrint(result.Status))
```
jlpadilla, line 92: -Use klog.
- Use a higher verbosity.

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+			authenticated, err := verifyToken(tokenId, r)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			if !authenticated {
+				http.Error(w, "Invalid token", http.StatusForbidden)
+				return
+			}
+			fmt.Fprintf(w, "Authentication successful!")
```
jlpadilla, line 74: - log the event, but don't respond.
- call `next()`
```suggestion
			log.V(5).Info("User authentication successful!")
			next()
```

pkg/rbac/middleware.go
jlpadilla: We can log it if you find it helpful for debugging problems later, but certainly use a higher verbosity level. **Hint**, you could separate it in 2 different logs with different verbosity. Example:
```go
klog.V(5).Info("Got user token from cookie.")
klog.V(9).Info("TokenReview result: %s", resultData)
```

main.go
Anxhela21: changed to gorilla mux

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		klog.Fatal(err.Error())
+	}
+	kClientset, err = kubernetes.NewForConfig(config)
+
+	if err != nil {
+		klog.Fatal(err.Error())
+	}
+
+}
```
jlpadilla, line 28: Could you move this block to `config/kubeClient.go` ?

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,97 @@
...
+
+//verifies token (userid) with the TokenReview:
+func Middleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			//if there is cookie available use that else use the authorization header:
+			var clientToken string
+			if len(r.Cookies()) > 0 {
+				for _, cookie := range r.Cookies() {
+					if cookie.Name == "acm-access-token-cookie" {
```
jlpadilla, line 38: You can access the cookie by name instead of iterating through all cookies.
https://pkg.go.dev/net/http#Request.Cookie

```go
cookie, err := r.Cookie("acm-access-token-cookie")
```

pkg/rbac/middleware.go
Anxhela21: Yes, and we can remove the kube call from main too and directly call when we authenticate with the token review in the middleware.

pkg/rbac/middleware.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/stolostron/search-v2-api/pkg/config"
+	authv1 "k8s.io/api/authentication/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+)
+
+//verifies token (userid) with the TokenReview:
+func Middleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
```
xrajesh, line 16: Confused by multiple returns here . Couldnt we have `func Middleware(next http.Handler) http.Handler {` @Anxhela21 @jlpadilla

Anxhela21: I don't think so because in server.go we call the middleware using `router.Use(...)`
(defined as: `func (*mux.Router).Use(mwf ...mux.MiddlewareFunc)`) which expects the middleware to be of type `mux.MiddlewareFunc` so if we call using: router.Use(rbac.Middleware(next)) to return http.Handler type we get an error.

jlpadilla: Not sure if this will work, but you can try this.
```
router.Use(rbac.Middleware)
```

Then here
```
func Middleware() func(next http.Handler) http.Handler {
```

Anxhela21: that should work as long as we return _a function returning the http.Handler_ instead of directly returning one with Middleware()

PR #34, opened 2022-05-03
Update relations query

SherinV:
https://github.com/stolostron/backlog/issues/15197
https://github.com/stolostron/backlog/issues/20553


pkg/resolver/search.go
```diff
@@ -196,121 +201,162 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
+	srcDestIds := make([]interface{}, 0)
+	srcDestIds = append(srcDestIds, goqu.I("e.sourceid"), goqu.I("e.destid"))
+
+	// Non-recursive term
+	baseTerm := goqu.From(schema.Table("all_edges").As("e")).
+		Select(selectBase...).
+		Where(goqu.ExOr{"sourceid": (s.uids), "destid": (s.uids)})
+
+	// Recursive term
+	recursiveTerm := goqu.From(schema.Table("all_edges").As("e")).
```
SherinV, line 246: This needs a PR to indexer to add the view definition of all_edges.

pkg/resolver/search.go
```diff
@@ -19,6 +20,8 @@ import (
 	"k8s.io/klog/v2"
 )
 
+var mux sync.RWMutex // Mutex to lock map during read/write
```
jlpadilla, line 23: This is global so it's shared by all requests.  I think we should declare it in the `SearchResult` struct so it is scoped to an individual request.

pkg/resolver/search.go
SherinV: Added in https://github.com/stolostron/search-indexer/pull/45

pkg/resolver/related.go
```diff
@@ -22,3 +31,274 @@ type SearchRelatedResult struct {
 // 	klog.Info("TODO: Resolve SearchRelatedResult: Items() - model/related.go")
 // 	return nil
 // }
+func (s *SearchResult) buildRelationsQuery() {
+	var whereDs []exp.Expression
+
+	//The level can be parameterized later, if needed
+	whereDs = append(whereDs, goqu.C("level").Lt(4))       // Add filter to select only upto level 4 relationships
```
Anxhela21, line 38: Using level 4 as default seems high. Is this needed for accessing subscriptions?

pkg/resolver/related.go
```diff
@@ -22,3 +31,274 @@ type SearchRelatedResult struct {
...
+
+	//Final select statement
+	selectFinal := make([]interface{}, 0)
+	selectFinal = append(selectFinal, goqu.C("iid"), goqu.C("kind"), goqu.MIN("level").As("level"))
+
+	//GROUPBY CLAUSE
+	groupBy := make([]interface{}, 0)
+	groupBy = append(groupBy, goqu.C("iid"), goqu.C("kind"))
+
+	// Original query to find relations between resources - accepts an array of uids
```
Anxhela21, line 65: I don't think we need the `Original` in `Original query...` anymore - this is just a part of current the query.

pkg/resolver/search.go
```diff
@@ -73,7 +75,9 @@ func (s *SearchResult) Related() []SearchRelatedResult {
 	var r []SearchRelatedResult
 
 	if len(s.uids) > 0 {
+		startTime := time.Now()
 		r = s.getRelations()
+		klog.Info("Time taken to resolve relationships: ", time.Since(startTime))
```
jlpadilla, line 80: **Maybe this could be a separate PR.** Can we use the same pattern from the indexer?
- Log if it takes more than the expected time.
- Log if v=4

pkg/resolver/related.go
```diff
@@ -22,3 +31,274 @@ type SearchRelatedResult struct {
...
+	baseTerm := goqu.From(schema.Table("all_edges").As("e")).
+		Select(selectBase...).
+		Where(goqu.ExOr{"sourceid": (s.uids), "destid": (s.uids)})
+
+	// Recursive term
+	recursiveTerm := goqu.From(schema.Table("all_edges").As("e")).
+		InnerJoin(goqu.T("search_graph").As("sg"),
+			goqu.On(goqu.ExOr{"sg.destid": srcDestIds, "sg.sourceid": srcDestIds})).
+		Select(selectNext...).
+		// Limiting upto level 4 as it should suffice for application relations
```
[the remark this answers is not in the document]
SherinV: For single cluster, in ms. But, this has to be tested at scale.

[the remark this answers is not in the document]
SherinV: For single cluster, in ms. But, this has to be tested at scale.

pkg/resolver/related.go
```diff
@@ -22,3 +32,307 @@ type SearchRelatedResult struct {
...
+	// For normal searches, we go only upto level 1. This can be changed later, if necessary.
+	level := config.Cfg.RelationLevel
+	//The level can be parameterized later, if needed
+	whereDs = append(whereDs, goqu.C("level").Lte(level))  // Add filter to select upto level (default 3) relationships
+	whereDs = append(whereDs, goqu.C("uid").NotIn(s.uids)) // Add filter to avoid selecting the search object itself
+
+	//Non-recursive term SELECT CLAUSE
+	schema := goqu.S("search")
+	selectBase := make([]interface{}, 0)
+	selectBase = append(selectBase, goqu.L("1").As("level"), "sourceid", "destid", "sourcekind", "destkind", "cluster")
```
jlpadilla, line 80: You can do this with 1 line, which can help improve the readability of this function.  There's several instances below.
```suggestion
	selectBase := []interface{}{ goqu.L("1").As("level"), "sourceid", "destid", "sourcekind", "destkind", "cluster"}
```

pkg/resolver/mocks/mock-rel-1.json
```diff
@@ -1,25 +1,15 @@
 {
-    "columns": ["data", "destid", "destkind"],
+    "columns": ["iid", "kind", "level"],
```
jlpadilla, line 2: I believe `iid` was changed to `uid`.

pkg/resolver/related.go
```diff
@@ -22,3 +32,301 @@ type SearchRelatedResult struct {
...
+	// -- union -- This is added if `kind:Cluster` is present in search term
+	// -- select uid as uid, data->>'kind' as kind, 1 AS "level" FROM search.resources where cluster IN ('local-cluster')
+
+	var whereDs []exp.Expression
+	// This level will come into effect only in case of Application/Subscription relations.
+	// For normal searches, we go only upto level 1. This can be changed later, if necessary.
+	level := config.Cfg.RelationLevel
+	//The level can be parameterized later, if needed
+	whereDs = append(whereDs, goqu.C("level").Lte(level))  // Add filter to select upto level (default 3) relationships
+	whereDs = append(whereDs, goqu.C("uid").NotIn(s.uids)) // Add filter to avoid selecting the search object itself
```
jlpadilla, line 75: You should be able to simplify this one too.
```suggestion
	// This level will come into effect only in case of Application/Subscription relations.
	// For normal searches, we go only up to level 1. This can be changed later, if necessary.
	level := config.Cfg.RelationLevel
	
	// The level can be parameterized later, if needed
	whereDs := []exp.Expression{
		goqu.C("level").Lte(level))     // Add filter to select up to level (default 3) relationships
		goqu.C("uid").NotIn(s.uids)) // Add filter to avoid selecting the search object itself
```

SherinV: Done

pkg/resolver/related.go
```diff
@@ -22,3 +32,302 @@ type SearchRelatedResult struct {
...
+	// GROUP BY "uid", "kind"
+	// -- union -- This is added if `kind:Cluster` is present in search term
+	// -- select uid as uid, data->>'kind' as kind, 1 AS "level" FROM search.resources where cluster IN ('local-cluster')
+
+	// This level will come into effect only in case of Application/Subscription relations.
+	// For normal searches, we go only upto level 1. This can be changed later, if necessary.
+	level := config.Cfg.RelationLevel
+	//The level can be parameterized later, if needed
+	// whereDs = append(whereDs, goqu.C("level").Lte(level))  // Add filter to select upto level (default 3) relationships
+	// whereDs = append(whereDs, goqu.C("uid").NotIn(s.uids)) // Add filter to avoid selecting the search object itself
```
jlpadilla, line 74: Delete this
```suggestion
```

pkg/resolver/related.go
SherinV: Done

pkg/resolver/mocks/mock-rel-1.json
SherinV: Done

pkg/resolver/related.go
SherinV: Done

pkg/resolver/related_readme.md
```diff
@@ -0,0 +1,63 @@
+**Search Term** = Resource(s) searched for from the search bar.
```
SherinV, line 1: @jlpadilla Added a small readme here. We can move it somewhere else, if it doesn't fit here, or remove it altogether, if it is covered elsewhere.

pkg/resolver/related.go
```diff
@@ -22,3 +32,328 @@ type SearchRelatedResult struct {
...
+	//defining variables
+	relatedMap := map[string][]string{} // Map to store relations
+
+	// Build the relations query
+	s.buildRelationsQuery()
+
+	relations, relQueryError := s.pool.Query(context.TODO(), s.query, s.params...) // how to deal with defaults.
+	if relQueryError != nil {
+		klog.Errorf("Error while executing getRelations query. Error :%s", relQueryError.Error())
+		return nil
```
jlpadilla, line 223: Maybe it's safer to return an empty array instead of nil. We had a similar bug where the nil value broke the UI.

SherinV: Done

pkg/resolver/related.go
```diff
@@ -22,3 +32,328 @@ type SearchRelatedResult struct {
...
+
+	// iterating through resulting rows and scaning data, destid  and destkind
+	for relations.Next() {
+		var kind, uid string
+		var level int
+		relatedResultError := relations.Scan(&uid, &kind, &level)
+
+		if relatedResultError != nil {
+			klog.Errorf("Error %s retrieving rows for relationships:%s", relatedResultError.Error(), relations)
+			return nil
```
jlpadilla: In this case we may have partial results.  In GraphQL we can return the partial results and report the error, but I don't know how to handle this with the go library.
Maybe logging the error on the server is good enough here because this error should never happen.

pkg/resolver/related.go
```diff
@@ -22,3 +32,327 @@ type SearchRelatedResult struct {
...
+	//iterating and sending values to relatedSearch
+	for kind, items := range relatedItems {
+		relatedSearch[i] = SearchRelatedResult{Kind: kind, Items: items}
+		i++
+	}
+	return relatedSearch
+}
+
+func (s *SearchResult) updateKindMap(uid string, kind string, levelMap map[string][]string) {
+	s.mux.RLock() // Lock map to read
```
jlpadilla, line 306: I wonder if the lock is really needed here. As far as I can tell everything runs in the same go routine, I don't see `go func()` anywhere.

This won't cause a problem, just wondering if it could be simplified.

pkg/resolver/related_readme.md
jlpadilla: Thanks!  Maybe we can add a link from the v2-operator wiki after we merge.
https://github.com/stolostron/search-v2-operator/wiki

pkg/resolver/related.go
SherinV: Done

pkg/resolver/related.go
SherinV: Removed

pkg/resolver/related.go
```diff
@@ -22,3 +32,327 @@ type SearchRelatedResult struct {
...
+	level := config.Cfg.RelationLevel
+	//The level can be parameterized later, if needed
+
+	//Set level
+	if s.searchApplication() && level == 0 {
+		level = 3 // If search involves applications and level is not explicitly set by user, set to 3
+	} else if level == 0 {
+		level = 1 // If level is not explicitly set by user, set to 1
+	}
+	return level
```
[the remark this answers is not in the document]
SherinV: Added level to searchResult struct to enable logging in Related function.

jlpadilla:
/lgtm

@SherinV Big THANK YOU for this amazing contribution!

PR #38, opened 2022-05-04
Remove properties starting with underscore(_) from searchSchema suggestions

jlpadilla:

Issue: https://github.com/stolostron/backlog/issues/22013

### Changes
Remove properties that start with underscore(_) from the searchSchema because these are internal and not meant to be exposed to end users.


SherinV (approved):

jlpadilla:
/unhold

Verified PR image in my test cluster.
image

PR #30, opened 2022-04-27
Document the Search API (GraphQL)

jlpadilla:
Signed-off-by: Jorge Padilla <jpadilla@redhat.com>

PR #41, opened 2022-05-13
updates for date and num filter

SherinV:

**Related Issue:**  [stolostron/backlog#20787](https://github.com/stolostron/backlog/issues/20787)

### Description of changes
Add searchComplete and search support for date and number filters

### DIFF
--- a/pkg/resolver/search.go
+++ b/pkg/resolver/search.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/doug-martin/goqu/v9"
 	"github.com/doug-martin/goqu/v9/exp"
@@ -86,8 +87,6 @@ func (s *SearchResult) Uids() {
 	s.resolveUids()
 }
 
-//=====================
-
 func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid bool) {
 	var limit int
 	var selectDs *goqu.SelectDataset
@@ -121,20 +120,21 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
 
 	//LIMIT CLAUSE
 	if !count {
-		if s.input != nil && s.input.Limit != nil && *s.input.Limit > 0 {
-			limit = *s.input.Limit
-		} else if s.input != nil && s.input.Limit != nil && *s.input.Limit == -1 {
-			klog.Warning("No limit set. Fetching all results.")
-		} else {
-			limit = config.Cfg.QueryLimit
-		}
+		limit = s.setLimit()
 	}
+	var params []interface{}
+	var sql string
+	var err error
 	//Get the query
-	sql, params, err := selectDs.Where(whereDs...).Limit(uint(limit)).ToSQL()
-	if err != nil {
-		klog.Errorf("Error building SearchComplete query: %s", err.Error())
+	if limit != 0 {
+		sql, params, err = selectDs.Where(whereDs...).Limit(uint(limit)).ToSQL()
+	} else {
+		sql, params, err = selectDs.Where(whereDs...).ToSQL()
 	}
-	klog.V(3).Infof("query: %s\nargs: %s", sql, params)
+	if err != nil {
+		klog.Errorf("Error building Search query: %s", err.Error())
+	}
+	klog.V(3).Infof("Search query: %s\nargs: %s", sql, params)
 	s.query = sql
 	s.params = params
 }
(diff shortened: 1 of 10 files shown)


jlpadilla:
I'm adding the following doc in my GraphQL schema PR. #30 

```graphql
input SearchFilter {
    """
    Name of the property (key).
    """
    property: String!
    """
    Values for the property. Multiple values per property are interpreted as an OR operation.
    Optionally one of these operands [=,!,!=,>,>=,<,<=] can be included at the begining of the value.
    By default the equality operand is used.
    """
    values: [String]!
  }
```

pkg/resolver/search.go
```diff
@@ -196,6 +198,88 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
+		"!=", "",
+		"!", "",
+		"<", "",
+		">", "",
+		"=", "")
+	newValues := make([]string, len(values)) // To store cleaned up values (without operator)
+	for i, value := range values {
+		operatorRemovedValue := replacer.Replace(value)
+		newValues[i] = operatorRemovedValue
+		operator = strings.Replace(value, operatorRemovedValue, "", 1) // find operator
```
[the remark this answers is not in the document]
jlpadilla: Multiple operands are somewhat working in V1, but that wasn't fully intended so we need to decide what to do here.

I tried this query in V1.
image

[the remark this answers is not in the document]
jlpadilla: @SherinV I don't want this discussion to derail the PR, everything else looks great.  If we don't have an easy solution we can add to our bugs list.

pkg/resolver/search.go
```diff
@@ -395,11 +479,21 @@ func pointerToStringArray(pointerArray []*string) []string {
 	return values
 }
 
+func stringArrayToPointer(stringArray []string) []*string {
```
jlpadilla, line 482: Do we need this function? It looks like this it's only referenced from the test file.

image

pkg/resolver/testHelper_test.go
```diff
@@ -131,35 +162,97 @@ func stringInSlice(a string, list []string) bool {
...
-					data := item.(map[string]interface{})["properties"].(map[string]interface{})
+				}
 
-					if !stringInSlice(data[filter.Property].(string), values) {
+				switch op {
+				case "<":
+					if int(data[filter.Property].(float64)) < filterValue {
+						return true
+					}
+					return false
```
jlpadilla, line 217: This is the same and more compact.

```suggestion
					return int(data[filter.Property].(float64)) < filterValue
```

pkg/resolver/search.go
```diff
@@ -196,6 +198,88 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
+	operator := ">"
+	for i, val := range values {
+		switch val {
+		case "hour":
+			then := now.Add(time.Duration(-1) * time.Hour).Format("2006-01-02T15:04:05Z")
+			newValues[i] = then
+
+		case "day":
+			then := now.AddDate(0, 0, -1).Format("2006-01-02T15:04:05Z")
+			newValues[i] = then
```
jlpadilla, line 261: We could reduce these to a single line.
```suggestion
			newValues[i] = now.AddDate(0, 0, -1).Format("2006-01-02T15:04:05Z")
```

pkg/resolver/search.go
SherinV: Moved to test file

pkg/resolver/search.go
SherinV: Updated

pkg/resolver/search.go
```diff
@@ -410,12 +528,22 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
-				if filter.Property == "cluster" {
-					whereDs = append(whereDs, goqu.C(filter.Property).In(values).Expression())
-				} else {
-					whereDs = append(whereDs, goqu.L(`"data"->>?`, filter.Property).In(values).Expression())
+				//Sort map according to keys - This is for the ease/stability of tests when there are multiple operators
+				keys := make([]string, 0, len(opDateValueMap))
+				for k := range opDateValueMap {
+					keys = append(keys, k)
+				}
+				sort.Strings(keys)
```
jlpadilla, line 539: Would it be easier to store in a `[][]string` instead of  `map[string][]string`?
The `[][]string` will preserve the original order.

jlpadilla (approved):
/lgtm
/hold

pkg/resolver/search.go
SherinV: Map will be easier when multiple operators are involved, to look up and check. But, we can address it later if the use case is different.

PR #48, opened 2022-06-03
Cache all cluster-scoped resources

Anxhela21:
**Related Issue:**  stolostron/backlog#22507

### Description of changes
We want to authorize each user request. The changes in this PR are to get and cache all cluster-scoped resources (the next step will be to authorize).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,8 +17,8 @@ var Cfg = new()
 // Define a Config type to hold our config properties.
 type Config struct {
 	API_SERVER_URL string // address for Kubernetes API Server
-	AuthCacheTTL   int    // Time-to-live of Authentication (TokenReview) cache in milliseconds.
-	SharedCacheTTL int    // Time-to-live Hub Cluster scoped resources cache in milliseconds.
+	AuthCacheTTL   int    // Time-to-live (milliseconds) of Authentication (TokenReview) cache.
+	SharedCacheTTL int    // Time-to-live (milliseconds) of common resources (shared across users) cache.
 	ContextPath    string
 	DBHost         string
 	DBName         string
@@ -39,8 +39,8 @@ func new() *Config {
 	// Simply put, the order of preference is env -> default values (from left to right)
 	conf := &Config{
 		API_SERVER_URL: getEnv("API_SERVER_URL", "https://kubernetes.default.svc"),
-		AuthCacheTTL:   getEnvAsInt("AUTH_CACHE_TTL", int(60000)),
-		SharedCacheTTL: getEnvAsInt("AUTHOR_CACHE_TTL", int(600000)),
+		AuthCacheTTL:   getEnvAsInt("AUTH_CACHE_TTL", int(60000)),    // 1 minute
+		SharedCacheTTL: getEnvAsInt("SHARED_CACHE_TTL", int(120000)), // 2 min (increase to 10min after implementation)
 		ContextPath:    getEnv("CONTEXT_PATH", "/searchapi"),
 		DBHost:         getEnv("DB_HOST", "localhost"),
 		DBName:         getEnv("DB_NAME", ""),
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -28,7 +28,7 @@ type Config struct {
 	HttpPort       int
 	PlaygroundMode bool // Enable the GraphQL Playground client.
 	QueryLimit     int  // The default LIMIT to use on queries. Client can override.
-
+	RelationLevel  int  // The number of levels/hops for finding relationships for a particular resource
 	// Placeholder for future use.
 	// QueryLoopLimit          int // number of queries handled at a time
 	// RBAC_INACTIVITY_TIMEOUT int
@@ -50,7 +50,9 @@ func new() *Config {
 		HttpPort:       getEnvAsInt("HTTP_PORT", 4010),
 		PlaygroundMode: getEnvAsBool("PLAYGROUND_MODE", false),
 		QueryLimit:     getEnvAsInt("QUERY_LIMIT", 10000),
-
+		//Setting default level to 0 to check if user has explicitly set this variable
+		// This will be updated to 1 for default searches and 3 for applications - unless set by the user
+		RelationLevel: getEnvAsInt("RELATION_LEVEL", 0),
 		// Placeholder for future use.
 		// QueryLoopLimit:          getEnvAsInt("QUERY_LOOP_LIMIT", 5000),
 		// RBAC_INACTIVITY_TIMEOUT: getEnvAsInt("RBAC_INACTIVITY_TIMEOUT", 600000), // 10 minutes
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -18,6 +18,7 @@ var Cfg = new()
 type Config struct {
 	API_SERVER_URL string // address for Kubernetes API Server
 	AuthCacheTTL   int    // Time-to-live of Authentication (TokenReview) cache in milliseconds.
+	SharedCacheTTL int    // Time-to-live Hub Cluster scoped resources cache in milliseconds.
 	ContextPath    string
 	DBHost         string
 	DBName         string
@@ -39,6 +40,7 @@ func new() *Config {
 	conf := &Config{
 		API_SERVER_URL: getEnv("API_SERVER_URL", "https://kubernetes.default.svc"),
 		AuthCacheTTL:   getEnvAsInt("AUTH_CACHE_TTL", int(60000)),
+		SharedCacheTTL: getEnvAsInt("AUTHOR_CACHE_TTL", int(600000)),
 		ContextPath:    getEnv("CONTEXT_PATH", "/searchapi"),
 		DBHost:         getEnv("DB_HOST", "localhost"),
 		DBName:         getEnv("DB_NAME", ""),
--- a/pkg/rbac/authenticationMiddleware.go
+++ b/pkg/rbac/authenticationMiddleware.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"strings"
 
+	db "github.com/stolostron/search-v2-api/pkg/database"
 	"k8s.io/klog/v2"
 )
 
@@ -45,5 +46,16 @@ func AuthenticateUser(next http.Handler) http.Handler {
 		klog.V(6).Info("User authentication successful!")
 		next.ServeHTTP(w, r)
 
+	})
+}
+
+func AuthorizeUser(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		//get user cluster-scoped resources and cache:
+		err := cache.getClusterScopedResources(db.GetConnection())
+		if err != nil {
+			klog.Warning("Unexpected error while obtaining cluster-scoped resources.", err)
+		}
 	})
 }
--- a/pkg/rbac/authenticationMiddleware.go
+++ b/pkg/rbac/authenticationMiddleware.go
@@ -2,6 +2,7 @@
 package rbac
 
 import (
+	"fmt"
 	"net/http"
 	"strings"
 
@@ -58,4 +59,5 @@ func AuthorizeUser(token string) {
 	if err != nil {
 		klog.Warning("Unexpected error while obtaining cluster-scoped resources.", err)
 	}
+	fmt.Println("Finished getting cluster-scoped resources. Now Authorizing..") //place-holder comment.
 }
--- a/pkg/rbac/authenticationMiddleware.go
+++ b/pkg/rbac/authenticationMiddleware.go
@@ -2,12 +2,16 @@
 package rbac
 
 import (
-	"fmt"
+	"context"
 	"net/http"
 	"strings"
 
 	"k8s.io/klog/v2"
 )
+
+type ContextKey string
+
+const ContextAuthTokenKey ContextKey = "authToken"
 
 // verifies token (userid) with the TokenReview:
 func AuthenticateUser(next http.Handler) http.Handler {
@@ -45,19 +49,21 @@ func AuthenticateUser(next http.Handler) http.Handler {
 
 		klog.V(6).Info("User authentication successful!")
 		//if we are authenticated can move on to next step: authorize: get cluster scopedresources and cache
-		AuthorizeUser(clientToken)
 
-		next.ServeHTTP(w, r)
+		ctx := context.WithValue(r.Context(), ContextAuthTokenKey, clientToken)
+
+		next.ServeHTTP(w, r.WithContext(ctx))
 
 	})
 }
 
-func AuthorizeUser(token string) {
+// func AuthorizeUser(token string) {
 
-	//get user cluster-scoped resources and cache:
-	err := cache.checkUserResources(token)
-	if err != nil {
-		klog.Warning("Unexpected error while obtaining cluster-scoped resources.", err)
-	}
-	fmt.Println("Finished getting cluster-scoped resources. Now Authorizing..") //place-holder comment.
-}
+// 	//get user cluster-scoped resources and cache:
+// 	clientToken := r.Context().Value(ContextAuthTokenKey).(string)
+// 	err := cache.checkUserResources(token)
+// 	if err != nil {
+// 		klog.Warning("Unexpected error while obtaining cluster-scoped resources.", err)
+// 	}
+// 	fmt.Println("Finished getting cluster-scoped resources. Now Authorizing..") //place-holder comment.
+// }
--- a/pkg/rbac/authenticationMiddleware.go
+++ b/pkg/rbac/authenticationMiddleware.go
@@ -35,7 +35,7 @@ func AuthenticateUser(next http.Handler) http.Handler {
 			return
 		}
 
-		authenticated, err := Instcache.IsValidToken(r.Context(), clientToken)
+		authenticated, err := cacheInst.IsValidToken(r.Context(), clientToken)
 		if err != nil {
 			klog.Warning("Unexpected error while authenticating the request token.", err)
 			http.Error(w, "{\"message\":\"Unexpected error while authenticating the request token.\"}", http.StatusInternalServerError)
--- a/pkg/rbac/authenticationMiddleware.go
+++ b/pkg/rbac/authenticationMiddleware.go
@@ -35,11 +35,12 @@ func AuthenticateUser(next http.Handler) http.Handler {
 			return
 		}
 
-		authenticated, err := cache.IsValidToken(r.Context(), clientToken)
+		authenticated, err := Instcache.IsValidToken(r.Context(), clientToken)
 		if err != nil {
 			klog.Warning("Unexpected error while authenticating the request token.", err)
 			http.Error(w, "{\"message\":\"Unexpected error while authenticating the request token.\"}", http.StatusInternalServerError)
 			return
+
 		}
 		if !authenticated {
 			klog.V(4).Info("Rejecting request: Invalid token.")
@@ -47,7 +48,7 @@ func AuthenticateUser(next http.Handler) http.Handler {
 			return
 		}
 
-		klog.V(6).Info("User authentication successful!")
+		klog.V(4).Info("User authentication successful!")
 		//if we are authenticated can move on to next step: authorize: get cluster scopedresources and cache
 
 		ctx := context.WithValue(r.Context(), ContextAuthTokenKey, clientToken)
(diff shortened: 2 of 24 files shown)


pkg/rbac/authenticationMiddleware.go
```diff
@@ -43,7 +44,20 @@ func AuthenticateUser(next http.Handler) http.Handler {
 		}
 
 		klog.V(6).Info("User authentication successful!")
+		//if we are authenticated can move on to next step: authorize: get cluster scopedresources and cache
+		AuthorizeUser(clientToken)
```
jlpadilla, line 48: `server.go` is a better place for calling `AuthorizeUser()`. This way we can keep each middleware function independent.

```go
	// Add authentication middleware to the /searchapi (ContextPath) subroute.
	apiSubrouter := router.PathPrefix(config.Cfg.ContextPath).Subrouter()
	apiSubrouter.Use(rbac.AuthenticateUser)
	// if we are authenticated can move on to next step: authorize: get cluster scopedresources and cache
	apiSubrouter.Use(rbac.AuthorizeUser)
	apiSubrouter.Handle("/graphql", srv.Handler)
```

Then, use the request context to share the `clienToken` across the different middleware functions.
Reference: https://stackoverflow.com/questions/39946583/how-to-pass-context-in-golang-request-to-middleware

```go
type ContextKey string
const ContextAuthTokenKey ContextKey = "authToken"


// Add clientToken to request context.
ctx := context.WithValue(r.Context(), ContextAuthTokenKey, clientToken)
// Continue executing the middleware chain.
next.ServeHTTP(w, r.WithContext(ctx))
```

Access the clientToken from AuthorizeUser()
```go
clientToken := r.Context().Value(ContextAuthTokenKey).(string)
```

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,66 @@
...
+		var kind, apigroup string
+		err := rows.Scan(&kind, &apigroup)
+		if err != nil {
+			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), query)
+		}
+
+		csrmap[kind] = apigroup
+	}
+	resourcelist := &clusterScopedResourcesList{resources: csrmap, err: err}
+	cache.clusterScoped[uid] = resourcelist
```
jlpadilla, line 63: This list is the same for all users. It represents all the resources available in the cluster.

The next step will be to check which of those resources the current user is authorized to access.

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,66 @@
...
+	defer rows.Close()
+	csrmap := make(map[string]string)
+	for rows.Next() {
+		var kind, apigroup string
+		err := rows.Scan(&kind, &apigroup)
+		if err != nil {
+			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), query)
+		}
+
+		csrmap[kind] = apigroup
```
jlpadilla, line 60: Multiple ApiGroups could define a Kind with the same name. With this structure we can only store a single `Kind 
+ ApiGroup` combination.  We need to redefine how we store the cluster resources list.

pkg/rbac/authenticationMiddleware.go
Anxhela21: this is what I had initially but figured we are only getting the resources not authorizing yet.

Anxhela21: and was not not sure how to pass token from middlewares but the above works thank you!

pkg/rbac/clusterScopedAuthorization.go
Anxhela21: got it. will replace that resourceList with the authorized resource list at the next step (that's why uid is the key)

pkg/rbac/authenticationMiddleware.go
```diff
@@ -43,7 +48,22 @@ func AuthenticateUser(next http.Handler) http.Handler {
...
+		//if we are authenticated can move on to next step: authorize: get cluster scopedresources and cache
+
+		ctx := context.WithValue(r.Context(), ContextAuthTokenKey, clientToken)
+
+		next.ServeHTTP(w, r.WithContext(ctx))
 
 	})
 }
+
+// func AuthorizeUser(token string) {
```
SherinV, line 60: delete commented code?

pkg/rbac/cache.go
```diff
@@ -4,20 +4,25 @@ package rbac
...
 
+	"github.com/jackc/pgx/v4/pgxpool"
+	// db "github.com/stolostron/search-v2-api/pkg/database"
 	v1 "k8s.io/client-go/kubernetes/typed/authentication/v1"
 )
 
 // Cache used to optimize requests to the Kubernetes API server.
 type Cache struct {
 	authClient          v1.AuthenticationV1Interface // This allows tests to replace with mock client.
+	pool                *pgxpool.Pool
```
jlpadilla, line 15: In the indexer we use `pgxpoolmock.PgxPool`.  [Link to code](https://github.com/stolostron/search-indexer/blob/main/pkg/database/connection.go#L21)

I believe this is because `pgxpoolmock` can accept a `pgxpool`, but the opposite didn't work.

pkg/config/config.go
```diff
@@ -18,6 +18,7 @@ var Cfg = new()
 type Config struct {
 	API_SERVER_URL string // address for Kubernetes API Server
 	AuthCacheTTL   int    // Time-to-live of Authentication (TokenReview) cache in milliseconds.
+	SharedCacheTTL int    // Time-to-live Hub Cluster scoped resources cache in milliseconds.
```
jlpadilla, line 21: small clarification 
```suggestion
	AuthCacheTTL   int    // Time-to-live (milliseconds) of Authentication (TokenReview) cache.
	SharedCacheTTL int    // Time-to-live (milliseconds) of common resources (shared across users) cache.
```

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,73 @@
...
+type sharedList struct {
+	err       error
+	resources map[string][]string
+	updatedAt time.Time
+}
+
+func (cache *Cache) checkUserResources(uid string) (bool, error) {
+
+	//check if we already have the resources in cache
+	cr, resourcesExist := cache.shared[string(uid)]
```
jlpadilla, line 22: Is this uid the user's UID? We shouldn't need it to access shared cache resources, but we'll need it for the cached user data.

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+type sharedList struct {
+	err       error
+	resources map[string][]string
+	updatedAt time.Time
+}
+
+func (cache *Cache) checkUserResources(uid string) (bool, error) {
+
```
jlpadilla, line 20: We need a lock here to ensure that only 1 process updates the cache.  Any other processes will wait here until the data is updated.

pkg/rbac/authenticationMiddleware.go
```diff
@@ -30,20 +35,25 @@ func AuthenticateUser(next http.Handler) http.Handler {
 			return
 		}
 
-		authenticated, err := cache.IsValidToken(r.Context(), clientToken)
+		authenticated, err := Instcache.IsValidToken(r.Context(), clientToken)
```
jlpadilla, line 38: Can you change to `cacheInst` to follow naming convention?
- Using uppercase makes it public outside of the package, we want to keep the cache instance private and only access through public functions.

(I also hit problems because of the name collision, this helps to avoid confusion)

pkg/rbac/authorizationMiddleware.go
```diff
@@ -0,0 +1,25 @@
...
+		//get user cluster-scoped resources and cache:
+		// clientToken := r.Context().Value(ContextAuthTokenKey).(string)
+		// uid := cache.tokenReviews[clientToken].tokenReview.UID
+		// _, err := Instcache.pool.Query(context.Background(), `Select Count(*) from search.resources`, nil)
+		// fmt.Println("Instacache pool works", err)
+		_, err := Instcache.checkUserResources()
+		if err != nil {
+			klog.Warning("Unexpected error while obtaining cluster-scoped resources.", err)
+		}
+		fmt.Println("Finished getting cluster-scoped resources. Now Authorizing..") //place-holder comment.
```
jlpadilla, line 22: Try using `klog.Info()`.  Makes it easier to later adjust with `.V(*)` as needed.

pkg/rbac/cache.go
```diff
@@ -13,11 +15,16 @@ type Cache struct {
 	tokenReviews        map[string]*tokenReviewResult
 	tokenReviewsPending map[string][]chan *tokenReviewResult
 	tokenReviewsLock    sync.Mutex
+	shared              sharedList
+	sharedLock          sync.Mutex
+	pool                pgxpoolmock.PgxPool
```
jlpadilla, line 20: This should be a pointer to the pool to ensure that we only have a single connection pool. We'll need to update `db.GetConnection()`
```suggestion
	pool                *pgxpoolmock.PgxPool
```

pkg/rbac/cache.go
```diff
@@ -13,11 +15,16 @@ type Cache struct {
 	tokenReviews        map[string]*tokenReviewResult
 	tokenReviewsPending map[string][]chan *tokenReviewResult
 	tokenReviewsLock    sync.Mutex
+	shared              sharedList
+	sharedLock          sync.Mutex
```
jlpadilla, line 19: This sharedLock can be moved to the `sharedList` struct because we don't need to lock at this level.

pkg/rbac/cache.go
Anxhela21: Issue with converting pgxpool -> pgxpoolmock in connection.go was `ConnectConfig() ` isn't part of pgxpoolmock package.
image

pkg/rbac/cache.go
Anxhela21: Ah, right!

pkg/rbac/clusterScopedResources_test.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/driftprogramming/pgxpoolmock"
+	"github.com/golang/mock/gomock"
+)
+
+// Initialize cache object to use tests.
+func newResourcesListCache(t *testing.T) (*pgxpoolmock.MockPgxPool, Cache) {
```
jlpadilla, line 12: - The `MockPgxPool` is included in the `Cache` so you only need to return the Cache here.
- I'd call it `mockResourcesListCache()` for clarity.
```suggestion
// Initialize a mock cache object to use in tests.
func mockResourcesListCache(t *testing.T) (Cache) {
```

pkg/rbac/clusterScopedResources_test.go
```diff
@@ -0,0 +1,112 @@
...
+	mockPool := pgxpoolmock.NewMockPgxPool(ctrl)
+	return mockPool, Cache{
+		shared: sharedList{},
+		pool:   mockPool,
+	}
+}
+
+func Test_getResources_emptyCache(t *testing.T) {
+
+	mockpool, cache := newResourcesListCache(t)
```
jlpadilla, line 24: I'd name this `mock_cache` for clarity and match the tokenReview tests.

pkg/rbac/clusterScopedResources_test.go
```diff
@@ -0,0 +1,112 @@
...
+	).Return(pgxRows, nil)
+
+	result, err := cache.checkUserResources()
+
+	if len(result.resources) != 0 {
+		t.Error("Resources not in cache.")
+	}
+	if err != nil {
+		t.Error("Unexpected error while obtaining cluster-scoped resources.", err)
+	}
```
jlpadilla, line 40: You can also validate that `updatedAt` has been updated.

```
// Verify that cache was updated within the last 1 millisecond.
if mock_cache.shared.updatedAt.Before(time.Now().Add(time.Duration(-1) * time.Millisecond)) {
  t.Error("Expected cache.shared.updatedAt to be less than 1 millisecond ago.")
}
}

File: pkg/rbac/clusterScopedResources_test.go [hunk 16]
```diff
@@ -0,0 +1,112 @@
...
+	resourcemap := make(map[string][]string)
+	var apigroups string
+	var kinds []string
+
+	kinds = append(kinds, "kind1", "kind2")
+	apigroups = "apigroup1"
+
+	resourcemap[apigroups] = kinds
+	cache.shared = sharedList{
+		updatedAt: time.Now().Add(time.Duration(-2) * time.Minute),
```
jlpadilla, line 99: I'd set this to a bigger difference like 3 mins to make it clear that it's expired and avoid possible test flakiness if everything runs within the same time unit.

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,78 @@
...
+		defer rows.Close()
+		csrmap := make(map[string][]string)
+
+		for rows.Next() {
+			var kind string
+			var apigroup string
+			err := rows.Scan(&apigroup, &kind)
+			if err != nil {
+				klog.Errorf("Error %s retrieving rows for query:%s for apigroup %s and kind %s", err.Error(), query, apigroup, kind)
+			}
```
jlpadilla, line 67: Errors are per row, so the result could have 3 rows and only 1 has an error.
If a row has an error we should skip adding it and save the error.  
```suggestion
			err := rows.Scan(&apigroup, &kind)
			if err != nil {
				klog.Errorf("Error %s retrieving rows for query:%s for apigroup %s and kind %s", err.Error(), query, apigroup, kind)
				sharedList.err = err // This will only keep the last row with an error. It's unlikely that we see an error here, so reporting the last one is "good enough". The log above will record all instances.
				continue // skip adding this row because it had errors.
			}
```

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,78 @@
...
+
+	//"SELECT DISTINCT(data->>apigroup, data->>kind) FROM search.resources WHERE
+	// cluster='local-cluster' AND namespace=NULL"
+	query, _, err := ds.SelectDistinct(goqu.COALESCE(goqu.L(`"data"->>'apigroup'`), "").As("apigroup"),
+		goqu.COALESCE(goqu.L(`"data"->>'kind'`), "").As("kind")).
+		Where(goqu.L(`"cluster"::TEXT = 'local-cluster'`), goqu.L(`"data"->>'namespace'`).IsNull()).ToSQL()
+	if err != nil {
+		klog.Errorf("Error creating query [%s]. Error: [%+v]", query, err)
+	}
+	rows, queryerr := cache.pool.Query(context.Background(), query)
```
jlpadilla, line 53: Ideally we'll use the context we get from the request here.  If that isn't easy, then use `context.TODO()` and we can revisit it later.

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,78 @@
...
+	if queryerr != nil {
+		klog.Errorf("Error resolving query [%s]. Error: [%+v]", query, queryerr.Error())
+	}
```
jlpadilla, line 56: I believe that an error here will prevent us from doing anything else.
```suggestion
	if queryerr != nil {
		klog.Errorf("Error resolving query [%s]. Error: [%+v]", query, queryerr.Error())
		sl.err = queryErr
		sl.resources = make(map[string][]string)
		//sl.updatedAt = time.Now() // <- We'll need to think how we go about caching errors we may need to add logic for retry.  Let's keep this PR simple and add the error/retry logic later.
		return sl.resources, sl.err
	}
```

pkg/rbac/clusterScopedResources_test.go
Anxhela21: This is related issue with replacing pgxpool too pgxpoolmock in connection.go.

Anxhela21: If I remove the returned pool, the cache passes pgxpoolmock instead of *pgxpoolmock (since in Cache.go we have pgxpoolmock)

jlpadilla: Ahh, ok got it.  Thanks for clarifying it!

pkg/rbac/clusterScopedResources_test.go
Anxhela21: Done.

pkg/rbac/clusterScopedResources_test.go
Anxhela21: Done.

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+
+			csrmap[apigroup] = append(csrmap[apigroup], kind)
+
+		}
+	}
+
+	// Then update the cache.
+	shared.resources = csrmap
+	shared.err = err
```
jlpadilla, line 78: Small fix on the error handling.
```suggestion
	shared.err = nil // Clear previous errors.
	
	// if cache data non-exist or expired query shared esources:
	schemaTable := goqu.S("search").Table("resources")
	ds := goqu.From(schemaTable)

	//"SELECT DISTINCT(data->>apigroup, data->>kind) FROM search.resources WHERE
	// cluster='local-cluster' AND namespace=NULL"
	query, _, err := ds.SelectDistinct(goqu.COALESCE(goqu.L(`"data"->>'apigroup'`), "").As("apigroup"),
		goqu.COALESCE(goqu.L(`"data"->>'kind'`), "").As("kind")).
		Where(goqu.L(`"cluster"::TEXT = 'local-cluster'`), goqu.L(`"data"->>'namespace'`).IsNull()).ToSQL()
	if err != nil {
		klog.Errorf("Error creating query [%s]. Error: [%+v]", query, err)
		shared.err = err
	}

	rows, queryerr := cache.pool.Query(ctx, query)
	if queryerr != nil {
		klog.Errorf("Error resolving query [%s]. Error: [%+v]", query, queryerr.Error())
		shared.err = queryerr
		shared.resources = map[string][]string{}
		return shared.resources, shared.err
	}
	csrmap := make(map[string][]string)
	if rows != nil {
		defer rows.Close()

		for rows.Next() {
			var apigroup, kind string
			err := rows.Scan(&apigroup, &kind)
			if err != nil {
				klog.Errorf("Error %s retrieving rows for query:%s for apigroup %s and kind %s", err.Error(), query, apigroup, kind)
				shared.err = err
				continue
			}

			csrmap[apigroup] = append(csrmap[apigroup], kind)

		}
	}

	// Then update the cache.
	shared.resources = csrmap
```

jlpadilla: Sorry it's hard to see the changes.
Added at line 37
```
shared.err = nil // Clear previous errors.
```
Added at line 50
```
shared.err = err
```
Removed last line
```
shared.err = err
```

pkg/rbac/clusterScopedResources_test.go
```diff
@@ -0,0 +1,125 @@
...
+	).Return(pgxRows, nil)
+
+	resourcemap := make(map[string][]string)
+	var apigroups string
+	var kinds []string
+
+	kinds = append(kinds, "kind1", "kind2")
+	apigroups = "apigroup1"
+
+	resourcemap[apigroups] = kinds
```
jlpadilla, line 70: Could reduce to 1 line. Same in the next test.
```suggestion
	resourcemap := map[string][]string{"apigroup1": []string{"kind1", "kind2"}}
```

pkg/rbac/authorizationMiddleware.go
```diff
@@ -0,0 +1,19 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		_, err := cacheInst.ClusterScopedResources(r.Context())
+		if err != nil {
+			klog.Warning("Unexpected error while obtaining cluster-scoped resources.", err)
+		}
+		klog.Info("Finished getting cluster-scoped resources. Now Authorizing..")
+
+	})
+}
```
jlpadilla, line 19: Should call next when done.
```suggestion
func AuthorizeUser(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {

		_, err := cacheInst.ClusterScopedResources(r.Context())
		if err != nil {
			klog.Warning("Unexpected error while obtaining cluster-scoped resources.", err)
		}
		klog.Info("Finished getting cluster-scoped resources. Now Authorizing..")
		
		next.ServeHTTP(w, r.WithContext(ctx))
	})
}
```

jlpadilla (approved):
/lgtm
Thanks @Anxhela21 !

PR #50, opened 2022-06-06
select distinct results for search

SherinV:

### Description of changes
Select distinct results during keyword search to avoid duplicate results when searching for kind:Pod search


jlpadilla (approved):
/lgtm

PR #52, opened 2022-06-07
ILIKE for case-insensitive comparison on KIND

SherinV:

### Description of changes
Handle case-insensitive comparison in kind

### DIFF
--- a/pkg/resolver/search.go
+++ b/pkg/resolver/search.go
@@ -14,6 +14,7 @@ import (
 	"github.com/doug-martin/goqu/v9"
 	"github.com/doug-martin/goqu/v9/exp"
 	"github.com/driftprogramming/pgxpoolmock"
+	"github.com/lib/pq"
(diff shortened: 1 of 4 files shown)


pkg/resolver/search.go
```diff
@@ -268,11 +269,12 @@ func getWhereClauseExpression(prop, operator string, values []string) []exp.Expr
 			exps = append(exps, goqu.L(`"data"->>?`, prop).Gt(val))
 		}
 	case "=":
-
 		exps = append(exps, goqu.L(`"data"->>?`, prop).In(values))
 	default:
 		if prop == "cluster" {
 			exps = append(exps, goqu.C(prop).In(values))
+		} else if prop == "kind" { //ILIKE to enable case-insensitive comparison for kind
```
jlpadilla, line 276:
```suggestion
		} else if prop == "kind" { //ILIKE to enable case-insensitive comparison for kind. Needed for V1 compatibility.
```

PR #62, opened 2022-06-29
Namespace-resources cache and authorization

Anxhela21:
**Related Issue:**  stolostron/backlog#22508 and  stolostron/backlog#22511

### Description of changes
Changed implementation of how we get namespaces to better align with desired outcome


pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -7,14 +7,17 @@ import (
...
-	err       error
-	resources map[string][]string
-	updatedAt time.Time
-	lock      sync.Mutex
+	err        error
+	resources  map[string][]string
+	namespaces []string
+	updatedAt  time.Time
+	lock       sync.Mutex
 }
```
jlpadilla, line 21: - I suggest that we rename this object (and file) now that it has a wider scope.
- I also did some renaming and organization, we'll need to update the rest of this file accordingly.
```go
// Cache data shared across all users.
type SharedData struct {
	// These are the data fields.
	csResources  map[string][]resource  // Cluster-scoped resources (ie. Node)
	namespaces   []string
	
	// These are internal objects to track the state of the cache.
	csErr         error        // Capture errors retrieving cluster-scoped resources.
	csLock        sync.Mutex   // Locks the csResources map while updating it.
	csUpdatedAt   time.Time    // Time when cluster-scoped data was last updated.
	nsErr         error        // Capture errors retrieving namespaces.
	nsLock        sync.Mutex   // Locks the namespaces array while updating it.
	nsUpdatedAt   time.Time    // Time when namespaces data was last updated.
}
```

This `resource` struct will be used here and also for namespaced resources, so maybe define it in a common file.
```go
// A Kubernetes resource is identified by its apigroup and kind.
// Resource version is not considered for RBAC.
type resource struct {
	apigroup.  string
	kind       string
}
```

pkg/rbac/cache.go
```diff
@@ -6,22 +6,33 @@ import (
...
+	corev1Client     corev1.CoreV1Interface
+	kubeClient       kubernetes.Interface
+	resConfig        *rest.Config
 	tokenReviews     map[string]*tokenReviewCache
 	tokenReviewsLock sync.Mutex
+	userLock         sync.Mutex
 	shared           clusterScopedResources
+	users            map[string]*userData // UID:{userdata} UID comes from tokenreview
 	pool             pgxpoolmock.PgxPool
 }
```
jlpadilla, line 28: Some reorganization and minor changes. 
- Use a pointer `*` if possible for the clients to avoid duplicating the object.
- Renamed `userLock` to `usersLock` for consistency.
- Renamed `resConfig` to `restConfig`

```go
// Cache used to minimize requests to external APIs (Kubernetes and Database)
type Cache struct {
	tokenReviews     map[string]*tokenReviewCache
	tokenReviewsLock sync.Mutex
	shared           clusterScopedResources
	users            map[string]*userData // UID:{userdata} UID comes from tokenreview
	usersLock        sync.Mutex
	
	// Clients to external APIs.
	authClient       *authv1.AuthenticationV1Interface
	corev1Client     *corev1.CoreV1Interface
	kubeClient       *kubernetes.Interface
	pool             *pgxpoolmock.PgxPool  // Database client
	restConfig       *rest.Config
}
```

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -77,7 +80,37 @@ func (shared *clusterScopedResources) getClusterScopedResources(cache *Cache, ct
...
+		allNamespaces = append(allNamespaces, cache.shared.namespaces...)
+
+	} else {
+
+		klog.V(5).Info("Getting namespaces from Kube Client..")
+
+		cache.resConfig = config.GetClientConfig()
+		clientset, err := kubernetes.NewForConfig(cache.resConfig)
+		if err != nil {
+			klog.Info("Error with creating a new clientset with impersonation config.", err.Error())
```
jlpadilla, line 96: - No impersonation here.
- Log as warning or error.
```suggestion
			klog.Warn("Error with creating a new clientset to get namespaces.", err.Error())
```
```suggestion
			klog.Info("Error with creating a new clientset with impersonation config.", err.Error())
```

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -77,7 +80,37 @@ func (shared *clusterScopedResources) getClusterScopedResources(cache *Cache, ct
 		}
 	}
 
+	//gather all namespaces in the cluster and cache in shared namespaces cache
```
jlpadilla, line 83: Can you separate the new logic below for namespaces to a different function?
It's better to organize the logic for namespaces and cluster-scoped resources as 2 separate things.

pkg/rbac/clusterScopedAuthorization.go
```diff
@@ -77,7 +80,37 @@ func (shared *clusterScopedResources) getClusterScopedResources(cache *Cache, ct
...
+	var allNamespaces []string
+	if len(cache.shared.namespaces) > 0 {
```
jlpadilla, line 85: Here also need to check that the cache is not expired.

pkg/rbac/namespacedAuthorization.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type userData struct {
+	// impersonate kubernetes.Interface //client with impersonation config
+	err        error
+	namespaces []string
+	// resources  map[string][]string //key:namespace value list of resources
+	updatedAt time.Time
+	lock      sync.Mutex
+}
```
jlpadilla, line 23: - Remove namespaces because we don't have a way to get an accurate list of namespaces for the user.
- Note that some of the fields aren't implemented with this PR. Keep those commented out for now.
 
```go
// Contains data about the resources the user is allowed to access.
type userData struct {
	// clusters     []string              // Managed clusters where the user has view access.
	// csResources  []resource            // Cluster-scoped resources on hub the user has list access.
	// nsResources  map[string][]resource // Namespaced resources on hub the user has list access. 
                                              //   key:namespace value list of resources
	
	// Internal fields to manage the cache.
	clustersErr       error      // Error while updating clusters data.
	clustersLock      sync.Mutex // Locks when clusters data is being updated.
	clustersUpdatedAt time.Time  // Time clusters was last updated.
	csrErr             error      // Error while updating cluster-scoped resources data.
	csrLock            sync.Mutex // Locks when cluster-scoped resources data is being updated.
	csrUpdatedAt       time.Time  // Time cluster-scoped resources was last updated.
	nsrErr             error      // Error while updating namespaced resources data.
	nsrLock            sync.Mutex // Locks when namespaced resources data is being updated.
	nsrUpdatedAt       time.Time  // Time namespaced resources was last updated.

	// impersonate *kubernetes.Interface // client with impersonation config
}
```

pkg/rbac/namespacedAuthorization.go
```diff
@@ -0,0 +1,90 @@
...
+
+	}
+	// create new instance
+	cache.users[uid] = &user
+	userData, err := user.getNamespaces(cache, ctx, clientToken)
+	return userData, err
+
+}
+
+func (user *userData) getNamespaces(cache *Cache, ctx context.Context, clientToken string) (*userData, error) {
```
jlpadilla, line 43: Logic in this function needs to be updated to get the namespaced-resources.
```
oc auth can-i --list -n <iterate-each-namespace>
```

pkg/rbac/clusterScopedAuthorization.go
Anxhela21: Done.

pkg/rbac/namespacedAuthorization.go
Anxhela21: We already define some of these in the sharedResources cache. Do we need to define them in both caches?

pkg/rbac/namespacedAuthorization.go
Anxhela21: Unclear what the ask is here. Isn't this the authorization part?

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+	ds := goqu.From(schemaTable)
+
+	//"SELECT DISTINCT(data->>apigroup, data->>kind) FROM search.resources WHERE
+	// cluster='local-cluster' AND namespace=NULL"
+	query, _, err := ds.SelectDistinct(goqu.COALESCE(goqu.L(`"data"->>'apigroup'`), "").As("apigroup"),
+		goqu.COALESCE(goqu.L(`"data"->>'kind'`), "").As("kind")).
+		Where(goqu.L(`"cluster"::TEXT = 'local-cluster'`), goqu.L(`"data"->>'namespace'`).IsNull()).ToSQL()
+	if err != nil {
+		klog.Errorf("Error creating query [%s]. Error: [%+v]", query, err)
+		shared.csErr = err
```
xrajesh, line 65: We need a return here @Anxhela21

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+
+		for rows.Next() {
+			var kind string
+			var apigroup string
+			err := rows.Scan(&apigroup, &kind)
+			if err != nil {
+				klog.Errorf("Error %s retrieving rows for query:%s for apigroup %s and kind %s", err.Error(), query,
+					apigroup, kind)
+				shared.csErr = err
+				continue
```
xrajesh, line 88: If we get error for one row, is this an error state ? - If we treat this as an error state we can break the loop - if we want to still use it , we have to remove the error right ?

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+	allNamespaces, nsErr := shared.GetSharedNamespaces(cache, ctx)
+	if nsErr != nil {
+		shared.nsErr = nsErr
+	}
+
+	// Then update the cache.
+	shared.namespaces = allNamespaces
+	shared.nsUpdatedAt = time.Now()
+
+	shared.csUpdatedAt = time.Now()
```
xrajesh, line 107: Do we need 2 times here ? both seem to have nearly same time ?

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+	if queryerr != nil {
+		klog.Errorf("Error resolving query [%s]. Error: [%+v]", query, queryerr.Error())
+		shared.csErr = queryerr
+		shared.csResources = []resource{}
+		return shared.csResources, shared.csErr
+	}
+
+	// var resource *resource
+	if rows != nil {
+		defer rows.Close()
```
xrajesh, line 78: Can we close this before we make a call to GetSharedNamespaces ?

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+
+			shared.csResources = append(shared.csResources, resource{apigroup: apigroup, kind: kind})
+
+		}
+	}
+
+	shared.nsLock.Lock()
+	defer shared.nsLock.Unlock()
+	allNamespaces, nsErr := shared.GetSharedNamespaces(cache, ctx)
```
xrajesh, line 98: Can we parallelize getClusterScopedResources and  GetSharedNamespaces - Need not be part of this PR .. but we can make a note @Anxhela21

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+
+	return shared.csResources, shared.csErr
+}
+
+//gather all namespaces in the cluster and cache in shared namespaces cache
+func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context) ([]string, error) {
+	var allNamespaces []string
+	if len(cache.shared.namespaces) > 0 {
+		klog.V(5).Info("Using namespaces from shared cache")
+		allNamespaces = append(allNamespaces, cache.shared.namespaces...)
```
xrajesh, line 117: We can return cache.shared.namespaces as is right , why append to empty array?

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+
+	} else {
+
+		klog.V(5).Info("Getting namespaces from Kube Client..")
+
+		cache.restConfig = config.GetClientConfig()
+		clientset, err := kubernetes.NewForConfig(cache.restConfig)
+		if err != nil {
+			klog.Warning("Error with creating a new clientset.", err.Error())
+		}
```
xrajesh, line 127: Getting clientset object cant be a separate function.

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+		namespaceList, kubeErr := clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+		if kubeErr != nil {
+			klog.Warning("Error resolving namespaces from KubeClient: ", kubeErr)
+			shared.nsErr = kubeErr
+			shared.namespaces = append(shared.namespaces, allNamespaces...)
```
xrajesh, line 132: Isnt it allNamespaces empty at this point , why do we append , we can return as is right ?

pkg/rbac/sharedData.go
```diff
@@ -0,0 +1,143 @@
...
+			return shared.namespaces, shared.nsErr
+		}
+
+		// add namespaces to allNamespace List
+		for _, n := range namespaceList.Items {
+			allNamespaces = append(allNamespaces, n.Name)
+		}
+
+	}
+	return allNamespaces, shared.nsErr
```
xrajesh, line 142: can we return `return allNamespaces, nil`

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+	cache.usersLock.Lock()
+	defer cache.usersLock.Unlock()
+	cachedUserData, userDataExists := cache.users[uid] //check if userData cache for user already exists
+	if userDataExists {
+		klog.V(5).Info("Using user data from cache.")
+		return cachedUserData, user.err
+
+	}
+	// create new instance
+	cache.users[uid] = &user
```
xrajesh, line 51: Looks like a logic error - we are assigning nil value , should the userData from next line . @Anxhela21

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+	nsrUpdatedAt time.Time  // Time namespaced resources was last updated.
+
+	// impersonate *kubernetes.Interface // client with impersonation config
+}
+
+func (cache *Cache) GetUserData(ctx context.Context, clientToken string) (*userData, error) {
+	var user userData
+	uid := cache.tokenReviews[clientToken].tokenReview.Status.User.UID //get uid from tokenreview
+	cache.usersLock.Lock()
+	defer cache.usersLock.Unlock()
```
xrajesh, line 43: Do we need the lock for looking up the cache , or we could live with locking only when we update the cache. ?

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+// The following achieves same result as oc auth can-i --list -n <iterate-each-namespace>
+func (user *userData) getNamespacedResources(cache *Cache, ctx context.Context, clientToken string) (*userData, error) {
+
+	//first we check if we already have user's namespaced resources in userData cache
+	user.nsrLock.Lock()
+	defer user.nsrLock.Unlock()
+	if len(user.nsResources) > 0 &&
+		time.Now().Before(user.nsrUpdatedAt.Add(time.Duration(config.Cfg.UserCacheTTL)*time.Millisecond)) {
+		klog.V(5).Info("Using user's namespaced resources from cache.")
+		return user, user.nsrErr
```
xrajesh, line 66: We need to return `return user, nil`

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+	// Internal fields to manage the cache.
+	// clustersErr       error      // Error while updating clusters data.
+	// clustersLock      sync.Mutex // Locks when clusters data is being updated.
+	// clustersUpdatedAt time.Time  // Time clusters was last updated.
+	err     error      // Error while getting user data from cache
+	csrErr  error      // Error while updating cluster-scoped resources data.
+	csrLock sync.Mutex // Locks when cluster-scoped resources data is being updated.
+	// csrUpdatedAt time.Time  // Time cluster-scoped resources was last updated.
+	nsrErr       error      // Error while updating namespaced resources data.
+	nsrLock      sync.Mutex // Locks when namespaced resources data is being updated.
```
xrajesh, line 33: Can we just use one lock for userData Object ?

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+		klog.V(5).Info("Using user's namespaced resources from cache.")
+		return user, user.nsrErr
+	}
+
+	// get all namespaces from shared cache:
+	klog.V(5).Info("Getting namespaces from shared cache.")
+	user.csrLock.Lock()
+	defer user.csrLock.Unlock()
+	allNamespaces := cache.shared.namespaces
+	// allNamespaces = removeDuplicateStr(allNamespaces)
```
xrajesh, line 74: we can return if len(allNamespaces) ==0

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+		//
+		rulesCheck := authzv1.SelfSubjectRulesReview{
+			Spec: authzv1.SelfSubjectRulesReviewSpec{
+				Namespace: ns,
+			},
+		}
+
+		result, err := impersClientset.SelfSubjectRulesReviews().Create(ctx, &rulesCheck, metav1.CreateOptions{})
+		if err != nil {
+			klog.Error("Error creating SelfSubjectRulesReviews ", err)
```
xrajesh, line 91: May want to add the namespace to the list of log values .

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		klog.V(9).Infof("TokenReview Kube API result: %v\n", prettyPrint(result.Status))
```
xrajesh, line 93: The above V(9) log , should that happen in the error case - we could have this as the else block

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+	if cache.authzClient == nil {
+
+		config.Impersonate = rest.ImpersonationConfig{
+			UserName: cache.tokenReviews[clientToken].tokenReview.Status.User.Username,
+			UID:      cache.tokenReviews[clientToken].tokenReview.Status.User.UID,
+		}
+
+		clientset, err := kubernetes.NewForConfig(cache.restConfig)
+		if err != nil {
+			klog.Error("Error with creating a new clientset with impersonation config.", err.Error())
```
xrajesh, line 125: its safe to return here , in case of error .

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+
+		result, err := impersClientset.SelfSubjectRulesReviews().Create(ctx, &rulesCheck, metav1.CreateOptions{})
+		if err != nil {
+			klog.Error("Error creating SelfSubjectRulesReviews ", err)
+		}
+		klog.V(9).Infof("TokenReview Kube API result: %v\n", prettyPrint(result.Status))
+
+		for _, rules := range result.Status.ResourceRules { //iterate objects
+			for _, verb := range rules.Verbs {
+				if verb == "list" { //drill down to list only
```
xrajesh, line 97: OR verb == "*"

pkg/rbac/userData.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Internal fields to manage the cache.
+	// clustersErr       error      // Error while updating clusters data.
+	// clustersLock      sync.Mutex // Locks when clusters data is being updated.
+	// clustersUpdatedAt time.Time  // Time clusters was last updated.
+	err     error      // Error while getting user data from cache
+	csrErr  error      // Error while updating cluster-scoped resources data.
+	csrLock sync.Mutex // Locks when cluster-scoped resources data is being updated.
+	// csrUpdatedAt time.Time  // Time cluster-scoped resources was last updated.
+	nsrErr       error      // Error while updating namespaced resources data.
```
xrajesh, line 32: This nsrErr is not used

pkg/rbac/sharedData.go
Anxhela21: Okay since we reset the error in line 85 `err := rows.Scan(&apigroup, &kind)` we can set `shared.csErr = err` if err != nil  - in line 92.

pkg/rbac/sharedData.go
Anxhela21: Done

pkg/rbac/sharedData.go
xrajesh: If we set the error - Do you still want to continue the loop OR ee can break ? 
Currently we can set an error , but we still populate the data from DB .

Anxhela21: the original idea was - if only one row has an error we don't want to exit the loop we still want to populate data and continue loop. Based on discussion, we can change the error to a log instead since we don't use that specific error for anything other than looking at the specific kind/apigroup that caused query to fail.

pkg/rbac/sharedData.go
Anxhela21: That's true.  We can cache the nsUpdatedAt inside the GetSharedNamespaces func.

pkg/rbac/sharedData.go
Anxhela21: Yes, we do not use the database to get the namespaces.

pkg/rbac/sharedData.go
Anxhela21: Yeah, we can optimize this with channels.

pkg/rbac/sharedData.go
Anxhela21: Yes, I think we changed the cache to include the `namespaces` field after implementing this part of the code - will remove that empty array and cache appropriately.

pkg/rbac/sharedData.go
Anxhela21: same as comment above - removed empty array

pkg/rbac/userData.go
Anxhela21: This lock here is to lock the cache while we are initializing a new user.

pkg/rbac/userData.go
Anxhela21: changed to the following:
```
user.nsrErr = nil
return user, user.nsrErr
```

pkg/rbac/userData.go
Anxhela21: we could but maybe it's good to keep two since they are used for different scenarios?

pkg/rbac/userData.go
Anxhela21: Done.

pkg/rbac/userData.go
Anxhela21: yeah, let's put that in else

pkg/rbac/userData.go
Anxhela21: Okay, I think we can use it for the if len(allNamespaces) == 0 logic from comment above.

pkg/rbac/sharedData.go
Anxhela21: Changed to:
```
shared.nsErr = nil
return shared.namespaces, shared.nsErr
```

xrajesh (approved):

PR #63, opened 2022-07-01
Reduce default limit to 1000

jlpadilla:

**Related Issue:**  stolostron/backlog#<ISSUE_NUMBER>

### DIFF
--- a/test/locust-users.py
+++ b/test/locust-users.py
@@ -13,7 +13,7 @@ class UserBehavior(TaskSet):
         self.do_post({
             "op": "searchByKeyword",
             "query": "query q($input: [SearchInput]) {\n  searchResult: search(input: $input) {\n    items\n    __typename\n  }\n}",
-            "variables": {"input":[{"keywords": ["prom"]}]}
+            "variables": {"input":[{"keywords": ["apiserver"]}], "limit": 1000}
         })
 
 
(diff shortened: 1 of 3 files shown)


xrajesh:
@jlpadilla @SherinV - since we touched all the queries , can we parameterize this , may be able to help with large clusters ?

jlpadilla:
> @jlpadilla @SherinV - since we touched all the queries , can we parameterize this , may be able to help with large clusters ?

@xrajesh This is parameterized with an env variable.  All the hardcoded values updated in this PR are in the test files.

test/locust-users.py
```diff
@@ -22,10 +22,9 @@ def do_filter_search(self):
         self.do_post({
             "op": "searchByFilter",
             "query": "query q($input: [SearchInput]) {\n  searchResult: search(input: $input) {\n    items\n    __typename\n  }\n}",
-            "variables": {"input":[{"filters": [{"property":"kind", "values":["Pod"]}]}]}
+            "variables": {"input":[{"filters": [{"property":"namespace", "values":["default"]}], "limit": 1000}]}
```
xrajesh, line 25: @jlpadilla @SherinV - do we need to classify the filters , for example kind:pod may match a lot of resources vs namespace:default .

xrajesh (approved):

test/locust-users.py
jlpadilla: Yes, but my motivation for this change is because we had to code a special case for `kind` to match case insensitive (`ILIKE`) to keep compatibility with V1.

PR #74, opened 2022-07-22
Managed clusters resources auth

Anxhela21:
**Related Issue:**  stolostron/backlog#21071

### Description of changes
Along with namespaced resources, we want to make sure our users are authorized to access resources on the managed clusters.


pkg/rbac/sharedData.go
```diff
@@ -145,3 +161,37 @@ func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context)
...
 }
+
+func (shared *SharedData) GetSharedManagedClusterNamespaces(cache *Cache, ctx context.Context) ([]string, error) {
+
+	if shared.managedClusters == nil ||
+		time.Now().After(shared.mcUpdatedAt.Add(time.Duration(config.Cfg.SharedCacheTTL)*time.Millisecond)) {
+
+		var managedClusters []string
+
+		var clusterVersionGvr = schema.GroupVersionResource{
```
SherinV, line 172: We can declare this as a constant

pkg/rbac/userData.go
```diff
@@ -154,7 +164,9 @@ func (user *userData) getNamespacedResources(cache *Cache, ctx context.Context,
 		klog.Warning("All namespaces array from shared cache is empty.", cache.shared.nsErr)
 		return user, cache.shared.nsErr
 	}
-	// allNamespaces = removeDuplicateStr(allNamespaces)
+
+	fmt.Println("All namespaces", allNamespaces)
```
SherinV, line 168: use klog instead

pkg/rbac/sharedData.go
```diff
@@ -145,3 +161,37 @@ func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context)
...
+			Group:    "cluster.open-cluster-management.io",
+			Version:  "v1",
+			Resource: "managedclusters",
+		}
+
+		cache.dynamicConfig = config.GetDynamicClient()
+
+		resourceObj, err := cache.dynamicConfig.Resource(clusterVersionGvr).List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			klog.Warning("Error resolving resources with dynamic client", err.Error())
```
SherinV, line 182: Do we have to return here?

pkg/rbac/userData.go
```diff
@@ -175,6 +175,8 @@ func (user *userData) getNamespacedResources(cache *Cache, ctx context.Context,
 		result, err := impersClientset.SelfSubjectRulesReviews().Create(ctx, &rulesCheck, metav1.CreateOptions{})
 		if err != nil {
 			klog.Error("Error creating SelfSubjectRulesReviews for namespace", err, ns)
+			user.nsrErr = err
+			return user, user.nsrErr
```
SherinV, line 179: Do we need to return the non-error namespaces here?

pkg/rbac/userData.go
```diff
@@ -192,7 +194,29 @@ func (user *userData) getNamespacedResources(cache *Cache, ctx context.Context,
 		}
 	}
 
+	//take intersection of authorizaed user namespaces and managed clusters:
+	managedClusters := cache.shared.managedClusters
+	namespaceNames := make([]string, len(nsResourcesAuth))
+	for name := range nsResourcesAuth {
+		namespaceNames = append(namespaceNames, name)
```
SherinV, line 201: It should be namespaceNames[i] = name. Since you already have assigned length, you should not append.

pkg/rbac/userData.go
```diff
@@ -192,7 +194,29 @@ func (user *userData) getNamespacedResources(cache *Cache, ctx context.Context,
...
+	}
+
+	namespaceInt, err := intersection(namespaceNames, managedClusters)
```
SherinV, line 204: This should be stored in user's cache as a list to store the managed clusters that user can access.

pkg/rbac/userData.go
```diff
@@ -192,7 +194,29 @@ func (user *userData) getNamespacedResources(cache *Cache, ctx context.Context,
...
+	}
+
+	nsResources := make(map[string][]resource)
+	for _, v := range namespaceInt {
+		if val, ok := nsResourcesAuth[v]; ok {
+			nsResources[v] = val
+		}
+	}
+
+	user.nsResources = nsResources
```
SherinV, line 217: Since this is for hub cluster namespaced resources, we shouldn't change this.

pkg/rbac/userData.go
SherinV: We should move it to a function on its own to populate user's managedClusters ([]string)

pkg/rbac/userData.go
xrajesh: we complete processing and send the namespaces that passed successfully .

pkg/rbac/userData.go
Anxhela21: Done.

pkg/rbac/userData.go
Anxhela21: Done.

pkg/rbac/sharedData.go
```diff
@@ -32,116 +39,164 @@ type resource struct {
...
+		if err == nil {
+			klog.V(6).Info("Successfully retrieved shared namespaces!")
+		}
+		// get all managed clustsers in cache
+		err = cache.shared.GetSharedManagedClusters(cache, ctx)
+		if err == nil {
+			klog.V(6).Info("Successfully retrieved managed clusters!")
+		}
+
+		return err
```
jlpadilla, line 71: These 3 actions should happen at the same time.
Here's an example of how to implement it using sync.WaitGroup.  https://gobyexample.com/waitgroups

pkg/rbac/sharedData.go
```diff
@@ -32,116 +39,164 @@ type resource struct {
...
+		return shared.csErr
+	}
 
+	rows, queryerr := cache.pool.Query(ctx, query)
+	if queryerr != nil {
+		klog.Errorf("Error resolving query [%s]. Error: [%+v]", query, queryerr.Error())
+		shared.csErr = queryerr
+		shared.csResources = []resource{}
+		return shared.csErr
+	}
```
jlpadilla, line 119: `shared.csUpdatedAt` is not updated in these error cases. We'll need to discuss how we'll handle this case.

- If we set `shared.csUpdatedAt` we won't be able to recover from the error until the cache expires.
- If we don't set `shared.csUpdatedAt` we might generate too many requests if we keep failing every time we retry to get this data.

We'll need something in between where errors are cached for a shorter time period.

pkg/rbac/sharedData.go
```diff
@@ -32,116 +39,164 @@ type resource struct {
...
+		(time.Now().Before(shared.nsUpdatedAt.Add(time.Duration(config.Cfg.SharedCacheTTL) * time.Millisecond))) &&
+		(time.Now().Before(shared.mcUpdatedAt.Add(time.Duration(config.Cfg.SharedCacheTTL) * time.Millisecond))) {
+
+		return true
+	}
+	return false
+}
 
-				shared.csResources = append(shared.csResources, resource{apigroup: apigroup, kind: kind})
+func (shared *SharedData) getClusterScopedResources(cache *Cache, ctx context.Context) error {
```
jlpadilla, line 88:
```suggestion
// Obtain all the cluster-scoped resources in the hub cluster that support list and watch.
// Equivalent to: `oc api-resources -o wide | grep false | grep watch | grep list`
func (shared *SharedData) getClusterScopedResources(cache *Cache, ctx context.Context) error {
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +39,164 @@ type resource struct {
...
+		}
 	}
+	shared.csUpdatedAt = time.Now()
 
-	return shared.csResources, shared.csErr
+	return shared.csErr
 }
 
-func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context) ([]string, error) {
+func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context) error {
```
jlpadilla, line 143:
```suggestion
// Obtain all the namespaces in the hub cluster.
// Equivalent to `oc get namespaces`
func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context) error {
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +39,164 @@ type resource struct {
...
-		shared.nsUpdatedAt = time.Now()
+	shared.mcLock.Lock()
+	defer shared.mcLock.Unlock()
+	// clear previous cache
+	shared.managedClusters = nil
+	shared.mcErr = nil
 
+	var managedClusters []string
+
+	resourceObj, err := cache.dynamicClient.Resource(clusterVersionGvr).List(context.TODO(), metav1.ListOptions{})
```
jlpadilla, line 186: We have a context, so let's use it.
```suggestion
	resourceObj, err := cache.dynamicClient.Resource(clusterVersionGvr).List(ctx, metav1.ListOptions{})
```

pkg/rbac/sharedData.go
jlpadilla: After thinking about this, this should work best:
- Here, update the timestamp to record that we made the API call.
- Then, in `sharedCacheValid()` we can decide how long an error response cache is valid.

Anxhela21: Okay, as discussed, will note this as optimization step in next user story.

pkg/rbac/sharedData.go
Anxhela21: Okay, as discussed, will note this as optimization step in next user story.

pkg/rbac/authzMiddleware.go
```diff
@@ -10,19 +10,23 @@ import (
...
 		}
 		klog.Info("Finished getting shared resources. Now getting user data..")
 
 		clientToken := r.Context().Value(ContextAuthTokenKey).(string)
-		_, newerr := cacheInst.GetUserData(r.Context(), clientToken, nil)
-		if newerr != nil {
-			klog.Warning("Unexpected error while obtaining user namespaces.", newerr)
+		_, userErr := cacheInst.GetUserData(r.Context(), clientToken, nil)
+		if userErr != nil {
+			klog.Warning("Unexpected error while obtaining user namespaces.", userErr)
```
jlpadilla, line 24:
```suggestion
			klog.Warning("Unexpected error while obtaining user data.", userErr)
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
+		if err == nil {
+			klog.V(6).Info("Successfully retrieved shared namespaces!")
+		}
+		// get all managed clustsers in cache
+		err = cache.shared.GetSharedManagedClusters(cache, ctx)
+		if err == nil {
+			klog.V(6).Info("Successfully retrieved managed clusters!")
+		}
+
+		return err
```
jlpadilla, line 72: This always returns the result of the last call. (GetSharedManagedClusters)
Errors in the previous calls won't be returned.

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
-	error) {
+func (cache *Cache) PopulateSharedCache(ctx context.Context) error {
+
+	if sharedCacheValid(&cache.shared) { //if all cache is valid we use cache data
+		klog.V(5).Info("Using shared data from cache.")
+		return nil
+	} else { //get data and cache
+
+		// get all cluster-scoped resources and cache in shared.csResources
+		err := cache.shared.getClusterScopedResources(cache, ctx)
```
jlpadilla, line 57: Use uppercase to match `GetSharedNamespaces` and `GetSharedManagedClusters`
```suggestion
		err := cache.shared.GetClusterScopedResources(cache, ctx)
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
+
+	if rows != nil {
+		defer rows.Close()
+
+		for rows.Next() {
+			var kind string
+			var apigroup string
+			err := rows.Scan(&apigroup, &kind)
+			if err != nil {
+				klog.Errorf("Error %s retrieving rows for query:%s for apigroup %s and kind %s", err.Error(), query,
```
jlpadilla, line 132: Lower to warning because this is a partial error.
```suggestion
				klog.Warnf("Error %s retrieving rows for query:%s for apigroup %s and kind %s", err.Error(), query,
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
+		shared.csResources = []resource{}
+		return shared.csErr
+	}
+
+	if rows != nil {
+		defer rows.Close()
+
+		for rows.Next() {
+			var kind string
+			var apigroup string
```
jlpadilla, line 129:
```suggestion
			var apigroup, kind string
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
+// Equivalent to `oc get namespaces`
+func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context) error {
+	shared.nsLock.Lock()
+	defer shared.nsLock.Unlock()
+	//empty previous cache
+	shared.namespaces = nil
+	shared.nsErr = nil
 
-				shared.csResources = append(shared.csResources, resource{apigroup: apigroup, kind: kind})
+	klog.V(5).Info("Getting namespaces from Kube Client..")
```
jlpadilla, line 155:
```suggestion
	klog.V(5).Info("Getting namespaces from Kube Client.")
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
+		shared.namespaces = append(shared.namespaces, n.Name)
 	}
+	shared.nsUpdatedAt = time.Now()
 
-	return shared.csResources, shared.csErr
+	return shared.nsErr
 }
 
-func (shared *SharedData) GetSharedNamespaces(cache *Cache, ctx context.Context) ([]string, error) {
+func (shared *SharedData) GetSharedManagedClusters(cache *Cache, ctx context.Context) error {
```
jlpadilla, line 182: No need to use `Shared` in the function name.
```suggestion
func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context) error {
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
-		clientset, err := kubernetes.NewForConfig(cache.restConfig)
-		if err != nil {
-			klog.Warning("Error with creating a new clientset.", err.Error())
+	resourceObj, err := cache.dynamicClient.Resource(managedClusterResourceGvr).List(ctx, metav1.ListOptions{})
 
-		}
+	if err != nil {
+		klog.Warning("Error resolving resources with dynamic client", err.Error())
+		return shared.mcErr
+	}
```
jlpadilla, line 200: Save the error and time in the cache.
```suggestion
	if err != nil {
		klog.Warning("Error resolving resources with dynamic client", err.Error())
		shared.mcErr = err
		shared.mcUpdatedAt = time.Now()
		return shared.mcErr
	}
```

pkg/rbac/sharedData.go
```diff
@@ -32,116 +40,173 @@ type resource struct {
...
 
-		// add namespaces to allNamespace List
-		for _, n := range namespaceList.Items {
-			shared.namespaces = append(shared.namespaces, n.Name)
-		}
-		shared.nsUpdatedAt = time.Now()
+	shared.managedClusters = managedClusters
+	shared.mcUpdatedAt = time.Now()
+	shared.mcErr = nil
```
jlpadilla, line 209: Remove this line. We did this at the beginning of this function.
```suggestion
```

pkg/rbac/userData.go
```diff
@@ -17,17 +17,24 @@ import (
...
-	// clusters     []string              // Managed clusters where the user has view access.
 	csResources []resource            // Cluster-scoped resources on hub the user has list access.
 	nsResources map[string][]resource // Namespaced resources on hub the user has list access.
+	clusters    []string              // Managed clusters where the user has view access.
 
+	clustersErr error // Error while updating clusters data.
+	// clustersLock      sync.Mutex // Locks when clusters data is being updated. NOTE: not implmented because we use the nsrLock
+	clustersUpdatedAt time.Time // Time clusters was last updated.
+
+	// Internal fields to manage the cache.
```
jlpadilla, line 28: Move comment.
```suggestion
	// Internal fields to manage the cache.
	clustersErr error // Error while updating clusters data.
	// clustersLock      sync.Mutex // Locks when clusters data is being updated. NOTE: not implmented because we use the nsrLock
	clustersUpdatedAt time.Time // Time clusters was last updated.
```

pkg/rbac/userData.go
```diff
@@ -176,23 +184,41 @@ func (user *userData) getNamespacedResources(cache *Cache, ctx context.Context,
...
+		for _, rules := range result.Status.ResourceRules {
 			for _, verb := range rules.Verbs {
-				if verb == "list" || verb == "*" { //drill down to list only
+				if verb == "list" || verb == "*" {
 					for _, res := range rules.Resources {
 						for _, api := range rules.APIGroups {
 							user.nsResources[ns] = append(user.nsResources[ns], resource{apigroup: api, kind: res})
 						}
 					}
 				}
```
jlpadilla, line 197: This will need to be optimized to account for `*`, but let's merge this PR first and update this right after.

pkg/rbac/sharedData.go
Anxhela21: Okay, maybe it's better to return if err not nil after each call.

pkg/rbac/sharedData.go
```diff
@@ -32,116 +39,171 @@ type resource struct {
...
+	klog.V(6).Info("Querying database for cluster-scoped resources.")
+
+	// Building query to get cluster scoped resources
+	// Original query: "SELECT DISTINCT(data->>apigroup, data->>kind) FROM search.resources WHERE
+	// cluster='local-cluster' AND namespace=NULL"
+	schemaTable := goqu.S("search").Table("resources")
+	ds := goqu.From(schemaTable)
+	query, _, err := ds.SelectDistinct(goqu.COALESCE(goqu.L(`"data"->>'apigroup'`), "").As("apigroup"),
+		goqu.COALESCE(goqu.L(`"data"->>'kind_plural'`), "").As("kind")).
+		Where(goqu.L(`"cluster"::TEXT = 'local-cluster'`), goqu.L(`"data"->>'namespace'`).IsNull()).ToSQL()
```
SherinV, line 113: From our discussion yesterday, use data->>'_hubClusterResource'='true' instead of "cluster"::TEXT = 'local-cluster' to avoid the usage of `local-cluster`

pkg/rbac/cache.go
```diff
@@ -34,5 +37,8 @@ var cacheInst = Cache{
 	usersLock:        sync.Mutex{},
 	shared:           SharedData{},
 	users:            map[string]*userData{},
+	restConfig:       config.GetClientConfig(),
 	pool:             db.GetConnection(),
```
SherinV, line 41: what happens if pool is nil here? Will we try again before sending queries?

Anxhela21: Good point. We can add a retry.

xrajesh (approved):

xrajesh:
Approving the test cases .

pkg/rbac/cache.go
SherinV: Tested this with @Anxhela21 - If postgres is down, when api comes up, the UI will never recover, since we will keep erroring out trying to get clusterscoped resources. Error is `nil pointer dereference`

PR #75, opened 2022-07-22
Xav/rbacmerged

xrajesh:
**Related Issue:**  stolostron/backlog#<ISSUE_NUMBER>


pkg/rbac/cache.go
```diff
@@ -25,12 +22,9 @@ type Cache struct {
 	// Clients to external APIs.
 	// Defining these here allow the tests to replace with a mock client.
 	authnClient  authnv1.AuthenticationV1Interface
-	authzClient  authzv1.AuthorizationV1Interface
 	corev1Client corev1.CoreV1Interface
-
-	kubeClient kubernetes.Interface
```
Anxhela21, line 31: Why do we remove the kubeClient from cache?

pkg/rbac/userData.go
```diff
@@ -18,41 +18,114 @@ import (
...
+	if err == nil {
+		klog.V(5).Info("No errors on namespacedresources not present ....Computing now", user)
+		userData, err = user.getClusterScopedResources(cache, ctx, clientToken)
+	}
 	return userData, err
 
 }
 
+func userCacheValid(user *userData) bool {
+	if (time.Now().Before(user.csrUpdatedAt.Add(time.Duration(config.Cfg.UserCacheTTL) * time.Millisecond))) &&
```
Anxhela21, line 66: Can we have a comment here or update the comments in user cache clarifying that these `updatedAt` fields are the time when resources were authorized last.

pkg/rbac/cache.go
xrajesh: It was not getting used.

pkg/rbac/userData.go
```diff
@@ -18,41 +18,113 @@ import (
...
+			user.authzClient = authzClient
+		}
 	}
-
 	userData, err := user.getNamespacedResources(cache, ctx, clientToken)
+
+	// Get cluster scoped resources for the user
+	// TO DO : Make this parallel operation
+	if err == nil {
+		klog.V(5).Info("No errors on namespacedresources not present ....Computing now", user)
```
Anxhela21, line 57:
```suggestion
		klog.V(5).Info("No errors on namespacedresources present ....Computing now", user)
```

Anxhela21: Also, do we need to log the user info? Maybe we can but at a higher verbosity?
```suggestion
		klog.V(6).Info("No errors on namespacedresources present ....Computing now", user)
```

xrajesh: If we want to log that , as well see the user too right ? Is this high priority - or can hold @Anxhela21

SherinV (approved):

PR #82, opened 2022-08-19
Add where clause for rbac

SherinV:

### Description of changes
- Adds RBAC WHERE clause to the search queries.

### DIFF
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -37,13 +37,23 @@ type UserData struct {
 	authzClient v1.AuthorizationV1Interface
 }
 
-func (cache *Cache) GetUserData(ctx context.Context, clientToken string,
+func (cache *Cache) GetUserData(ctx context.Context,
 	authzClient v1.AuthorizationV1Interface) (*UserData, error) {
 	var user *UserData
-	uid := cache.tokenReviews[clientToken].tokenReview.Status.User.UID //get uid from tokenreview
+	var uid string
+	clientToken := ctx.Value(ContextAuthTokenKey).(string)
+
+	//get uid from tokenreview
+	if tokenReview, err := cache.GetTokenReview(ctx, clientToken); err == nil {
+		uid = tokenReview.Status.User.UID
+	} else {
+		return user, err
+	}
+
 	cache.usersLock.Lock()
 	defer cache.usersLock.Unlock()
 	cachedUserData, userDataExists := cache.users[uid] //check if userData cache for user already exists
+
 	// UserDataExists and its valid
 	if userDataExists && userCacheValid(cachedUserData) {
 		klog.V(5).Info("Using user data from cache.")
@@ -244,14 +254,14 @@ func (user *UserData) getImpersonationClientSet(clientToken string, cache *Cache
 	return user.authzClient, nil
 }
 
-func (userdata *UserData) GetCsResources() []Resource {
-	return userdata.csResources
+func (user *UserData) GetCsResources() []Resource {
+	return user.csResources
 }
 
-func (userdata *UserData) GetNsResources() map[string][]Resource {
-	return userdata.nsResources
+func (user *UserData) GetNsResources() map[string][]Resource {
+	return user.nsResources
 }
 
-func (userdata *UserData) GetManagedClusters() []string {
-	return userdata.managedClusters
+func (user *UserData) GetManagedClusters() []string {
+	return user.managedClusters
 }
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -124,7 +124,8 @@ func (user *UserData) getClusterScopedResources(cache *Cache, ctx context.Contex
 	user.userResourceAccess.CsResources = nil
 	for _, res := range clusterScopedResources {
 		if user.userAuthorizedListCSResource(ctx, impersClientset, res.Apigroup, res.Kind) {
-			user.userResourceAccess.CsResources = append(user.userResourceAccess.CsResources, Resource{Apigroup: res.Apigroup, Kind: res.Kind})
+			user.userResourceAccess.CsResources = append(user.userResourceAccess.CsResources,
+				Resource{Apigroup: res.Apigroup, Kind: res.Kind})
 		}
 	}
 	user.csrUpdatedAt = time.Now()
@@ -207,7 +208,8 @@ func (user *UserData) getNamespacedResources(cache *Cache, ctx context.Context, 
 				if verb == "list" || verb == "*" { //TODO: resourceName == "*" && verb == "*" then exit loop
 					for _, res := range rules.Resources {
 						for _, api := range rules.APIGroups {
-							user.userResourceAccess.NsResources[ns] = append(user.userResourceAccess.NsResources[ns], Resource{Apigroup: api, Kind: res})
+							user.userResourceAccess.NsResources[ns] = append(user.userResourceAccess.NsResources[ns],
+								Resource{Apigroup: api, Kind: res})
 						}
 					}
 				}
@@ -219,7 +221,8 @@ func (user *UserData) getNamespacedResources(cache *Cache, ctx context.Context, 
 						if res == "managedclusterviews" {
 							for i := range managedClusters {
 								if managedClusters[i] == ns {
-									user.userResourceAccess.ManagedClusters = append(user.userResourceAccess.ManagedClusters, ns)
+									user.userResourceAccess.ManagedClusters =
+										append(user.userResourceAccess.ManagedClusters, ns)
 
 								}
 							}
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -16,8 +16,8 @@ import (
 )
 
 // Contains data about the resources the user is allowed to access.
-type UserData struct {
-	userResourceAccess UserDataCache
+type UserDataCache struct {
+	userData UserData
 
 	// Internal fields to manage the cache.
 	clustersErr error // Error while updating clusters data.
@@ -36,15 +36,15 @@ type UserData struct {
 }
 
 // Stuct to keep a copy of users access
-type UserDataCache struct {
+type UserData struct {
 	CsResources     []Resource            // Cluster-scoped resources on hub the user has list access.
 	NsResources     map[string][]Resource // Namespaced resources on hub the user has list access.
 	ManagedClusters []string              // Managed clusters where the user has view access.
 }
 
 func (cache *Cache) GetUserData(ctx context.Context,
-	authzClient v1.AuthorizationV1Interface) (*UserData, error) {
-	var user *UserData
+	authzClient v1.AuthorizationV1Interface) (*UserDataCache, error) {
+	var user *UserDataCache
 	var uid string
 	clientToken := ctx.Value(ContextAuthTokenKey).(string)
 
@@ -65,7 +65,7 @@ func (cache *Cache) GetUserData(ctx context.Context,
 		return cachedUserData, nil
 	} else {
 		// User not in cache , Initialize and assign to the UID
-		user = &UserData{}
+		user = &UserDataCache{}
 		cache.users[uid] = user
 		// We want to setup the client if passed, this is only for unit tests
 		if authzClient != nil {
@@ -89,7 +89,7 @@ func (cache *Cache) GetUserData(ctx context.Context,
 
 /* Cache is Valid if the csrUpdatedAt and nsrUpdatedAt times are before the
 Cache expiry time */
-func userCacheValid(user *UserData) bool {
+func userCacheValid(user *UserDataCache) bool {
 	if (time.Now().Before(user.csrUpdatedAt.Add(time.Duration(config.Cfg.UserCacheTTL) * time.Millisecond))) &&
 		(time.Now().Before(user.nsrUpdatedAt.Add(time.Duration(config.Cfg.UserCacheTTL) * time.Millisecond))) &&
 		(time.Now().Before(user.clustersUpdatedAt.Add(time.Duration(config.Cfg.UserCacheTTL) * time.Millisecond))) {
@@ -99,8 +99,8 @@ func userCacheValid(user *UserData) bool {
 }
 
 // Equivalent to: oc auth can-i list <resource> --as=<user>
-func (user *UserData) getClusterScopedResources(cache *Cache, ctx context.Context,
-	clientToken string) (*UserData, error) {
+func (user *UserDataCache) getClusterScopedResources(cache *Cache, ctx context.Context,
+	clientToken string) (*UserDataCache, error) {
 
 	// get all cluster scoped from shared cache:
 	klog.V(5).Info("Getting cluster scoped resources from shared cache.")
@@ -121,10 +121,10 @@ func (user *UserData) getClusterScopedResources(cache *Cache, ctx context.Contex
 		return user, user.csrErr
 	}
 	//If we have a new set of authorized list for the user reset the previous one
-	user.userResourceAccess.CsResources = nil
+	user.userData.CsResources = nil
 	for _, res := range clusterScopedResources {
 		if user.userAuthorizedListCSResource(ctx, impersClientset, res.Apigroup, res.Kind) {
-			user.userResourceAccess.CsResources = append(user.userResourceAccess.CsResources,
+			user.userData.CsResources = append(user.userData.CsResources,
 				Resource{Apigroup: res.Apigroup, Kind: res.Kind})
 		}
 	}
@@ -132,7 +132,7 @@ func (user *UserData) getClusterScopedResources(cache *Cache, ctx context.Contex
 	return user, user.csrErr
 }
 
-func (user *UserData) userAuthorizedListCSResource(ctx context.Context, authzClient v1.AuthorizationV1Interface,
+func (user *UserDataCache) userAuthorizedListCSResource(ctx context.Context, authzClient v1.AuthorizationV1Interface,
 	apigroup string, kind_plural string) bool {
 	accessCheck := &authz.SelfSubjectAccessReview{
 		Spec: authz.SelfSubjectAccessReviewSpec{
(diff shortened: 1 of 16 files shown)


pkg/rbac/userData.go
```diff
@@ -243,3 +243,15 @@ func (user *userData) getImpersonationClientSet(clientToken string, cache *Cache
...
+	return userdata.csResources
+}
+
+func (userdata *UserData) GetNsResources() map[string][]Resource {
+	return userdata.nsResources
+}
+
+func (userdata *UserData) GetManagedClusters() []string {
+	return userdata.managedClusters
+}
```
jlpadilla, line 257: Use `user` to stay consistent with the other methods.
```suggestion
func (user *UserData) GetCsResources() []Resource {
	return user.csResources
}

func (user *UserData) GetNsResources() map[string][]Resource {
	return user.nsResources
}

func (user *UserData) GetManagedClusters() []string {
	return user.managedClusters
}
```

pkg/resolver/search.go
```diff
@@ -107,6 +123,86 @@ func (s *SearchResult) Uids() {
 	s.resolveUids()
 }
 
+// Get a copy of the current user access if user data exists
+func userdataExists(ctx context.Context) (*UserResourceAccess, error) {
```
jlpadilla, line 127: I'd rename to `getUserAccessData()`.  Return an error if the data doesn't exist.

pkg/resolver/search.go
```diff
@@ -107,6 +123,86 @@ func (s *SearchResult) Uids() {
...
+		uid = tr.Status.User.UID
+	}
+
+	userData, userDataExistsErr := rbac.CacheInst.GetUserData(ctx, uid, nil)
+	useraccess := UserResourceAccess{}
+	if userDataExistsErr == nil {
+		useraccess = UserResourceAccess{CsResources: userData.GetCsResources(),
+			NsResources: userData.GetNsResources(), ManagedClusters: userData.GetManagedClusters()}
+	}
+	return &useraccess, userDataExistsErr
```
jlpadilla, line 141:
```suggestion
	userData, userDataErr := rbac.CacheInst.GetUserData(ctx, uid, nil)
	if userDataErr != nil {
		return nil, userDataErr
	}
	userAccess := $UserResourceAccess{
		CsResources: userData.GetCsResources(),
		NsResources: userData.GetNsResources(),
		ManagedClusters: userData.GetManagedClusters()}
	}
	return userAccess, userDataErr
```

pkg/resolver/search.go
```diff
@@ -21,9 +21,17 @@ import (
...
+	"github.com/stolostron/search-v2-api/pkg/rbac"
 	"k8s.io/klog/v2"
 )
 
+// Stuct to keep a copy of users access
+type UserResourceAccess struct {
+	CsResources     []rbac.Resource            // Cluster-scoped resources on hub the user has list access.
+	NsResources     map[string][]rbac.Resource // Namespaced resources on hub the user has list access.
+	ManagedClusters []string                   // Managed clusters where the user has view access.
+}
```
jlpadilla, line 33: Should move this definition to `userData.go` ?  
It can help to stay in sync with the `UserData` struct.

jlpadilla: Actually, we should use inheritance (or composition?) to represent these 2 structs better. I haven't used it before, so we don't have an example, but it should be something like this...

```go
type UserAccessData struct {
	CsResources     []rbac.Resource            // Cluster-scoped resources on hub the user has list access.
	NsResources     map[string][]rbac.Resource // Namespaced resources on hub the user has list access.
	ManagedClusters []string                   // Managed clusters where the user has view access.
}

type UserDataCache struct {
	UserAccessData

	// Add all the fields to manage the cache.
	// ...
}
```

pkg/resolver/search.go
```diff
@@ -136,6 +232,9 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
 	//WHERE CLAUSE
 	if s.input != nil && (len(s.input.Filters) > 0 || (s.input.Keywords != nil && len(s.input.Keywords) > 0)) {
 		whereDs = WhereClauseFilter(s.input)
+		if s.userAccess != nil {
+			whereDs = append(whereDs, buildRbacWhereClause(ctx, s.userAccess)) // add rbac
+		}
```
jlpadilla, line 237:
```go
} else {
    panic("RBAC clause is required!")
}
```
Well `panic` might be a little extreme, but it could help during development.  😄 
All queries must include the RBAC filter or return an error.

pkg/resolver/search.go
```diff
@@ -467,9 +566,11 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
 }
 
 func getKeys(stringArrayMap map[string][]string) []string {
-	keys := make([]string, 0, len(stringArrayMap))
+	i := 0
+	keys := make([]string, len(stringArrayMap))
 	for k := range stringArrayMap {
-		keys = append(keys, k)
+		keys[i] = k
+		i++
```
jlpadilla, line 573:
```golang
keys := make([]string, len(stringArrayMap))
for k, i := range stringArrayMap {
	keys[i] = k
}

File: pkg/resolver/search.go [hunk 7]
```diff
@@ -107,6 +123,86 @@ func (s *SearchResult) Uids() {
...
+	if userDataExistsErr == nil {
+		useraccess = UserResourceAccess{CsResources: userData.GetCsResources(),
+			NsResources: userData.GetNsResources(), ManagedClusters: userData.GetManagedClusters()}
+	}
+	return &useraccess, userDataExistsErr
+}
+
+// Build where clause with rbac by combining clusterscoped, namespace scoped and managed cluster access
+func buildRbacWhereClause(ctx context.Context, userrbac *UserResourceAccess) exp.ExpressionList {
+
```
jlpadilla, line 146: Maybe this is simpler to read?

```go
.Where(
    matchSearchQuery(),
    goqu.Or(
        matchManagedCluster()           // goqu.I("cluster").In([]string{"clusterNames", ....})
        goqu.And(
            matchHubCluster()           // goqu.L(`data->>?`, "_hubClusterResource").Eq("true")
            goqu.Or(
                matchClusterScopedResources(), // (namespace=null AND apigroup AND kind)
                matchNamespacedResources()     // (namespace AND apiproup AND kind)
            )
       )
   )
)
```

pkg/resolver/search.go
SherinV: But `i` is not an int in this case, right?

jlpadilla: 🤦 sorry, I didn't realize it's a map.

pkg/resolver/search.go
SherinV: function removed.

pkg/rbac/userData.go
SherinV: done

pkg/rbac/userData_test.go
```diff
@@ -394,7 +399,8 @@ func Test_managedClusters_emptyCache(t *testing.T) {
 		return true, notCreateRule, nil
 
 	})
-	result, err := mock_cache.GetUserData(context.Background(), "123456", fs.AuthorizationV1())
+	ctx := context.WithValue(context.TODO(), ContextAuthTokenKey, "123456")
```
jlpadilla, line 402: Actually `context.Background()` is better for the tests.  It yields the same and it doesn't imply that we should change it later.

pkg/resolver/search.go
```diff
@@ -107,6 +131,79 @@ func (s *SearchResult) Uids() {
...
+				matchClusterScopedResources(userrbac.CsResources), // (namespace=null AND apigroup AND kind)
+				matchNamespacedResources(userrbac.NsResources),    // (namespace AND apiproup AND kind)
+			),
+		),
+	)
+}
+
+// inner function to loop through resources and build the where clause
+func loopThroughResources(resources []rbac.Resource) exp.ExpressionList {
+	whereCsDs := make([]exp.ExpressionList, 1) // Stores the combined where clause for cluster scoped resources
```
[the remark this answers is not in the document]
SherinV: Updated

pkg/rbac/userData_test.go
SherinV: Updated

pkg/resolver/search.go
SherinV: moved

pkg/rbac/userData.go
```diff
@@ -37,6 +35,13 @@ type UserData struct {
 	authzClient v1.AuthorizationV1Interface
 }
 
+// Stuct to keep a copy of users access
+type UserResourceAccess struct {
```
jlpadilla, line 39: Would it be simpler to rename these structs as `UserData` and `UserDataCache`?

pkg/resolver/search.go
```diff
@@ -26,11 +26,11 @@ import (
...
-type UserResourceAccess struct {
-	CsResources     []rbac.Resource            // Cluster-scoped resources on hub the user has list access.
-	NsResources     map[string][]rbac.Resource // Namespaced resources on hub the user has list access.
-	ManagedClusters []string                   // Managed clusters where the user has view access.
-}
+// type UserResourceAccess struct {
+// 	CsResources     []rbac.Resource            // Cluster-scoped resources on hub the user has list access.
+// 	NsResources     map[string][]rbac.Resource // Namespaced resources on hub the user has list access.
+// 	ManagedClusters []string                   // Managed clusters where the user has view access.
+// }
```
jlpadilla, line 33: delete

pkg/resolver/search.go
```diff
@@ -147,7 +147,8 @@ func buildRbacWhereClause(ctx context.Context, userrbac *UserResourceAccess) exp
 
 // inner function to loop through resources and build the where clause
 func loopThroughResources(resources []rbac.Resource) exp.ExpressionList {
```
jlpadilla, line 149:
```suggestion
// inner function to loop through resources and build the where clause
// Resolves to something similar to:
//    ((apigroup='' AND kind='') OR (apigroup='' AND kind='') OR ... )
func matchApigroupKind(resources []rbac.Resource) exp.ExpressionList {
```

pkg/resolver/search.go
```diff
@@ -107,6 +131,80 @@ func (s *SearchResult) Uids() {
...
+func matchClusterScopedResources(csRes []rbac.Resource) exp.ExpressionList {
+	var whereNsDs exp.ExpressionList
+
+	if len(csRes) > 0 {
+		return goqu.And(goqu.COALESCE(goqu.L(`data->>?`, "namespace"), "").Eq(""),
+			loopThroughResources(csRes))
+
+	}
+	return whereNsDs
+}
```
jlpadilla, line 177:
```suggestion
func matchClusterScopedResources(csRes []rbac.Resource) exp.ExpressionList {
	if len(csRes) > 0 {
		return goqu.And(goqu.COALESCE(goqu.L(`data->>?`, "namespace"), "").Eq(""),
			loopThroughResources(csRes))
	}
	return exp.ExpressionList{}
}
```

pkg/resolver/search.go
```diff
@@ -467,9 +571,11 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
 }
 
 func getKeys(stringArrayMap map[string][]string) []string {
```
jlpadilla, line 573: You could make this more generic and use it from `matchNamespacedResources()`
```suggestion
func getKeys(stringArrayMap map[string]interface{}) []string {
```

pkg/resolver/search.go
```diff
@@ -107,6 +131,80 @@ func (s *SearchResult) Uids() {
...
+func matchHubCluster() exp.BooleanExpression {
+	//hub cluster
+	return goqu.L(`data->>?`, "_hubClusterResource").Eq("true")
+}
+
+func matchManagedCluster(managedClusters []string) exp.BooleanExpression {
+	//managed clusters
+	return goqu.C("cluster").Eq(goqu.Any(pq.Array(managedClusters)))
+}
+
```
jlpadilla, line 207: I'd consider moving the rbac functions to a new file because this one is getting big.

pkg/resolver/search.go
SherinV: Uodated function name and comment

pkg/resolver/search.go
SherinV: Tried this, but it threw some error. But, found another way.

pkg/rbac/userData.go
SherinV: Done

pkg/resolver/search.go
SherinV: requires changes in other places where this function is used.

pkg/resolver/search.go
SherinV: Tried that before - it is not working. But, made a similar change.

jlpadilla: thanks!

pkg/resolver/rbacHelper.go
```diff
@@ -0,0 +1,87 @@
...
+		if i == 0 {
+			whereCsDs = goqu.And(whereOrDs...) // First time, AND all conditions
+		} else {
+			//Next time onwards, perform OR with the existing conditions
+			whereCsDs = goqu.Or(whereCsDs, goqu.And(whereOrDs...))
+		}
+	}
+	return whereCsDs
+}
+func matchClusterScopedResources(csRes []rbac.Resource) exp.ExpressionList {
```
jlpadilla, line 35:
```suggestion
// Match cluster-scoped resources, which are identified by not having the namespace property.
// Resolves to something like:
//   (AND data->>'namespace' = '')
func matchClusterScopedResources(csRes []rbac.Resource) exp.ExpressionList {
```

pkg/resolver/rbacHelper.go
```diff
@@ -0,0 +1,87 @@
...
+	if len(csRes) > 0 {
+		return goqu.And(goqu.COALESCE(goqu.L(`data->>?`, "namespace"), "").Eq(""),
+			matchApigroupKind(csRes))
+
+	}
+	return exp.NewExpressionList(0, nil)
+}
+
+func matchNamespacedResources(nsResources map[string][]rbac.Resource) exp.ExpressionList {
```
jlpadilla, line 44:
```suggestion
// For each namespace, match the authorized resources (apigroup + kind)
// Resolves to some similar to:
//    (namespace = 'a' AND ((apigroup='' AND kind='') OR (apigroup='' AND kind='') OR ... ) OR
//    (namespace = 'b' AND ( ... ) OR (namespace = 'c' AND ( ... ) OR ...
func matchNamespacedResources(nsResources map[string][]rbac.Resource) exp.ExpressionList {
```

pkg/resolver/rbacHelper.go
```diff
@@ -0,0 +1,87 @@
...
+		sort.Strings(namespaces) //to make unit tests pass
+		for nsCount, namespace := range namespaces {
+			whereNsDs[nsCount] = goqu.And(goqu.L(`data->>?`, "namespace").Eq(namespace),
+				matchApigroupKind(nsResources[namespace]))
+		}
+	}
+	return goqu.Or(whereNsDs...)
+}
+
+func matchHubCluster() exp.BooleanExpression {
```
jlpadilla, line 63:
```suggestion
// Match resources in the hub. These are identified by containing the property _hubClusterResource=true
// Resolves to:
//    (data->>'_hubClusterResource' = true)
func matchHubCluster() exp.BooleanExpression {
```

pkg/resolver/rbacHelper.go
```diff
@@ -0,0 +1,87 @@
...
+	//hub cluster
+	return goqu.L(`data->>?`, "_hubClusterResource").Eq("true")
+}
+
+func matchManagedCluster(managedClusters []string) exp.BooleanExpression {
```
jlpadilla, line 68:
```suggestion
// Match resources from the managed clusters.
// Resolves to:
//    ( cluster IN ['a', 'b', ...] )
func matchManagedCluster(managedClusters []string) exp.BooleanExpression {
```

pkg/resolver/related_test.go
```diff
@@ -21,11 +22,13 @@ func Test_SearchResolver_Relationships(t *testing.T) {
 
 	// //take the uids from above as input
 	searchInput := &model.SearchInput{Filters: []*model.SearchFilter{{Property: "uid", Values: resultList}}}
-	resolver, mockPool2 := newMockSearchResolver(t, searchInput, resultList)
+	ura := rbac.UserDataCache{}
```
jlpadilla, line 25: Please update `ura` to match the new name.

pkg/resolver/search.go
```diff
@@ -33,20 +34,28 @@ type SearchResult struct {
...
+	userAccess *rbac.UserDataCache
 }
 
 func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, error) {
 	// For each input, create a SearchResult resolver.
 	srchResult := make([]*SearchResult, len(input))
+	userAccess, userDataErr := getUserAccessData(ctx)
+	if userDataErr != nil {
+		return nil, userDataErr
+	}
```
jlpadilla, line 46: 1. Let's log this error here. I't not being logged inside the function.
2. I think this error is included in the API query response, so we may need to convert into a more generic error like: "Unable to resolve query because of error while resolving user's access."
3. Maybe we should return the empty `srchResult` instead of nil. Will need to compare with V1.

SherinV: Is it okay to log the error inside the function? So, other resolvers like schema and searchComplete can use it too?

jlpadilla:
yes, we can log inside the function. Maybe we can also switch it to a generic message in the common function.

pkg/rbac/userData.go
```diff
@@ -37,20 +35,37 @@ type userData struct {
...
-func (cache *Cache) GetUserData(ctx context.Context, clientToken string,
-	authzClient v1.AuthorizationV1Interface) (*userData, error) {
-	var user *userData
-	uid := cache.tokenReviews[clientToken].tokenReview.Status.User.UID //get uid from tokenreview
+// Stuct to keep a copy of users access
+type UserDataCache struct {
+	CsResources     []Resource            // Cluster-scoped resources on hub the user has list access.
+	NsResources     map[string][]Resource // Namespaced resources on hub the user has list access.
+	ManagedClusters []string              // Managed clusters where the user has view access.
+}
```
jlpadilla, line 43: Naming is backwards.  I expect this struct to be `UserData` and the other one `UserDataCache` because it has the added fields t manage the caching.

pkg/resolver/rbacHelper.go
```diff
@@ -0,0 +1,104 @@
+package resolver
```
jlpadilla, line 1:
```suggestion
// Copyright Contributors to the Open Cluster Management project

package resolver
```

jlpadilla (approved):
/lgtm
/hold

Thank you @SherinV !

PR #83, opened 2022-08-29
update rbac for searchschema and searchcomplete

SherinV:

### Description of changes
Added  a map to lookup cluster scoped resources - to exclude them from namespace scoped filter.
Updated schema and searchcomplete with rbac clause.


pkg/rbac/userData.go
```diff
@@ -128,6 +129,7 @@ func (user *UserDataCache) getClusterScopedResources(cache *Cache, ctx context.C
 				Resource{Apigroup: res.Apigroup, Kind: res.Kind})
 		}
 	}
+	klog.V(7).Info("User has access to these cluster scoped res: ", user.userData.CsResources)
```
jlpadilla, line 132: Include the user id in case we need to debug multiple users.

SherinV: Added

SherinV:
PR under review

pkg/rbac/sharedData.go
```diff
@@ -17,6 +17,7 @@ import (
 type SharedData struct {
 	// These are the data fields.
 	csResources     []Resource // Cluster-scoped resources (ie. Node, ManagedCluster)
+	csResourcesMap  map[Resource]struct{}
```
jlpadilla, line 20: Wonder if we should avoid duplication and remove `csResources` and only use `csResourceMap` in all places.

pkg/rbac/sharedData.go
```diff
@@ -100,6 +101,7 @@ func (shared *SharedData) GetClusterScopedResources(cache *Cache, ctx context.Co
 	defer shared.csLock.Unlock()
 	//clear previous cache
 	shared.csResources = nil
+	shared.csResourcesMap = make(map[Resource]struct{})
```
jlpadilla, line 104: Try to keep these consistent, use either `nil or make(...)` on both to avoid confusion.
Would be better if we can only use the map.

pkg/rbac/userData.go
```diff
@@ -101,7 +114,11 @@ func userCacheValid(user *UserDataCache) bool {
 // Equivalent to: oc auth can-i list <resource> --as=<user>
 func (user *UserDataCache) getClusterScopedResources(cache *Cache, ctx context.Context,
 	clientToken string) (*UserDataCache, error) {
-
+	var uid string
+	var err error
+	if uid, err = cache.GetUserUID(ctx); err != nil {
+		klog.Warning("Error getting user's uid: ", err)
```
jlpadilla, line 120: Are we able to proceed without the `uid` or should we return an error here?

pkg/rbac/userData.go
```diff
@@ -235,13 +259,27 @@ func (user *UserDataCache) getNamespacedResources(cache *Cache, ctx context.Cont
...
+	klog.V(7).Infof("User %s has access to these ManagedClusters: %+v \n", uid, user.userData.ManagedClusters)
 
 	user.nsrUpdatedAt = time.Now()
 	user.clustersUpdatedAt = time.Now()
 
 	return user, user.nsrErr
 }
 
+//SSRR has resources that are clusterscoped too
+func (shared *SharedData) isClusterScoped(kindPlural, apigroup string) bool {
```
jlpadilla, line 272: 🤔 I wonder if it is possible to create a cluster-scoped CR with the same group and name as a namespaced CR.
This would definitely be an edge scenario.

pkg/resolver/searchComplete.go
```diff
@@ -74,6 +81,13 @@ func (s *SearchCompleteResult) searchCompleteQuery(ctx context.Context) {
 			//Adding notNull clause to filter out NULL values and ORDER by sort results
 			whereDs = append(whereDs, goqu.L(`"data"->>?`, s.property).IsNotNull())
 		}
+		//RBAC CLAUSE
+		if s.userData != nil {
+			whereDs = append(whereDs,
+				buildRbacWhereClause(ctx, s.userData)) // add rbac
+		} else {
+			panic("RBAC clause is required!")
```
jlpadilla, line 89: Do we have any context that we could add here to help us debug if we panic here? I'm thinking something like the user ID, the input query, or anything else that could help us later.

pkg/resolver/searchComplete.go
```diff
@@ -122,30 +136,32 @@ func (s *SearchCompleteResult) searchCompleteResults(ctx context.Context) ([]*st
...
-			srchCompleteOut = append(srchCompleteOutNum, srchCompleteOut[0], srchCompleteOut[len(srchCompleteOut)-1])
-		} else {
-			srchCompleteOut = append(srchCompleteOutNum, srchCompleteOut...)
-		}
+	if len(srchCompleteOut) > 0 {
+		//Check if results are date or number
+		isNumber := isNumber(srchCompleteOut)
+		if isNumber { //check if valid number
+			isNumber := "isNumber"
+			srchCompleteOutNum := []*string{&isNumber} //isNumber should be the first argument if the property is a number
```
jlpadilla, line 144: I got confused by the scope of `isNumber`, could you use different variable names?
```suggestion
		isNumber := isNumber(srchCompleteOut)
		if isNumber { //check if valid number
			isNumberStr := "isNumber"
			srchCompleteOutNum := []*string{&isNumberStr} //isNumber should be the first argument if the property is a number
```

pkg/resolver/searchComplete.go
```diff
@@ -122,30 +136,32 @@ func (s *SearchCompleteResult) searchCompleteResults(ctx context.Context) ([]*st
...
+				srchCompleteOut = append(srchCompleteOutNum, srchCompleteOut[0], srchCompleteOut[len(srchCompleteOut)-1])
+			} else {
+				srchCompleteOut = append(srchCompleteOutNum, srchCompleteOut...)
+			}
 
+		}
+		if !isNumber && isDate(srchCompleteOut) { //check if valid date
+			isDate := "isDate"
+			srchCompleteOutNum := []*string{&isDate}
+			srchCompleteOut = srchCompleteOutNum
```
jlpadilla, line 162: And please variables to match my previous comment.
```suggestion
		if !isNumber && isDate(srchCompleteOut) { //check if valid date
			isDateStr := "isDate"
			srchCompleteOutDate := []*string{&isDateStr}
			srchCompleteOut = srchCompleteOutDate
```

pkg/rbac/userData.go
SherinV: This uid is being used for logging only. Moved it.

pkg/rbac/sharedData.go
SherinV: Will move in next PR

pkg/resolver/searchComplete.go
```diff
@@ -92,9 +111,17 @@ func (s *SearchCompleteResult) searchCompleteQuery(ctx context.Context) {
...
 			klog.Errorf("Error building SearchComplete query: %s", err.Error())
 		}
+		f, err := os.Create("srchCompQuery.sql")
+		if err != nil {
+			fmt.Println("err creating file")
+		}
+		_, err = f.WriteString(sql)
+		if err != nil {
+			fmt.Println("err writing file")
+		}
```
jlpadilla, line 121: This looks like debug code. We could use a flag like `DEBUG_QUERY` if this can be helpful in the future.

pkg/resolver/searchComplete.go
```diff
@@ -116,35 +143,43 @@ func (s *SearchCompleteResult) searchCompleteResults(ctx context.Context) ([]*st
 		for rows.Next() {
 			prop := ""
 			scanErr := rows.Scan(&prop)
+			fmt.Println("prop:", prop)
```
jlpadilla, line 146: Please change to klog or delete.

jlpadilla (approved):
/lgtm
/hold

Please remove or adjust debugging code and logs before merging.

PR #84, opened 2022-08-29
Remove local-hub from managedCluster cache

Anxhela21:

### Description of changes
ManagedCluster cache in sharedData should not contain local-cluster.

### DIFF
--- a/pkg/rbac/sharedData.go
+++ b/pkg/rbac/sharedData.go
@@ -198,8 +198,9 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context) 
 	}
 
 	for _, item := range resourceObj.Items {
-		managedClusters = append(managedClusters, item.GetName())
-
+		if item.GetName() != "local-cluster" {
+			managedClusters = append(managedClusters, item.GetName())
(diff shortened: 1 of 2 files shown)


pkg/rbac/sharedData.go
```diff
@@ -198,8 +198,9 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context)
 	}
 
 	for _, item := range resourceObj.Items {
-		managedClusters = append(managedClusters, item.GetName())
-
+		if item.GetName() != "local-cluster" {
```
jlpadilla, line 201: This works for now, but I wonder if there's any way to avoid the dependency on the "local-cluster" name.

FYI: There's an RFE open to allow users to rename it.  I've heard this requirement several times before and it came up at today's architecture call again.

jlpadilla (approved):
/lgtm
/hold

pkg/rbac/sharedData.go
Anxhela21: That's a good point. Will merge for now and add improvement to the umbrella story.

PR #81, opened 2022-08-17
[WIP] process array props

SherinV:
Signed-off-by: Sherin Varughese <shvarugh@redhat.com>

**Related Issue:**  stolostron/backlog#<ISSUE_NUMBER>

PR #87, opened 2022-09-06
Array properties (search by label)

Anxhela21:
**Related Issue:**  stolostron/backlog#22013 &  stolostron/backlog#25645

### Description of changes

- Implements search by label.
- Autocomplete for label and other properties that support list.
- Not working yet, other list properties like container and image.

~~Continuing from PR https://github.com/stolostron/search-v2-api/pull/81~~


pkg/resolver/searchComplete.go
```diff
@@ -132,14 +135,63 @@ func (s *SearchCompleteResult) searchCompleteResults(ctx context.Context) ([]*st
...
+
+			case map[string]interface{}:
+				// fmt.Println("prop type: array")
+
+				arrayProperties[s.property] = struct{}{}
+				// klog.Info("arrayProperties: ", arrayProperties)
+				// fmt.Println("prop type: map[string]interface")
+				// labelStrings := make([]string, 0, len(v))
+				for key, value := range v {
+					labelString := fmt.Sprintf("%v=%v", key, fmt.Sprintf("%v", value))
```
jlpadilla, line 168: This should be the same.
```suggestion
					labelString := fmt.Sprintf("%s=%s", key, value))
```

pkg/resolver/search.go
```diff
@@ -395,7 +410,18 @@ func getOperatorAndNumDateFilter(values []string) map[string][]string {
 				then = now.AddDate(-1, 0, 0).Format(format)
 
 			default:
-				operator = ""
+				if filter == "label" {
+					klog.V(7).Info("filter is label. Operator is @>.")
+					operator = "@>"
```
jlpadilla, line 415: We need to be able to discover which properties are a list.

In v1 we hardcoded the list of properties, which isn't ideal. [Here's the v1 implementation.](https://github.com/stolostron/search-api/blob/main/src/v2/connectors/redisGraph.js#L25-L27)

Anxhela21: Done.

pkg/resolver/search.go
```diff
@@ -395,7 +411,21 @@ func getOperatorAndNumDateFilter(values []string) map[string][]string {
 				then = now.AddDate(-1, 0, 0).Format(format)
 
 			default:
-				operator = ""
+				//check that property value is an array:
+				array := strings.Split(string(val), ":")
```
jlpadilla, line 415: Why is the separator `:` ?

pkg/resolver/search.go
```diff
@@ -495,13 +526,29 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
+				if _, ok := arrayProperties[filter.Property]; ok {
+					cleanedVal := make([]string, len(values))
+					for i, val := range values {
+						labels := strings.Split(val, "=")
+						if len(labels) > 1 {
+							cleanedVal[i] = fmt.Sprintf(`{"%s":"%s"}`, labels[0], labels[1])
+						} else {
+							//// If property is of array type, format it as an array for easy searching
+							cleanedVal[i] = labels[0]
+						}
```
jlpadilla, line 540: A string like `label=a=b` will cause problems here. Let's update this to at least log this problem.
```suggestion
						if len(labels) == 2 {
							cleanedVal[i] = fmt.Sprintf(`{"%s":"%s"}`, labels[0], labels[1])
						} else if len(labels) == 1{
							//// If property is of array type, format it as an array for easy searching
							cleanedVal[i] = labels[0]
						} else {
							klog.error("Error while decoding label string")
							cleanedVal[i] = val // <<<REVIEW NOTE: is this the best we can do in this situation?
						}
```

pkg/resolver/searchComplete.go
```diff
@@ -78,9 +81,11 @@ func (s *SearchCompleteResult) searchCompleteQuery(ctx context.Context) {
 			whereDs = append(whereDs, goqu.C(s.property).IsNotNull(),
 				goqu.C(s.property).Neq("")) // remove empty strings from results
 		} else {
-			selectDs = ds.Select(goqu.L(`"data"->>?`, s.property).As("prop"))
+			//get data as jsoon object ->
```
jlpadilla, line 84:
```suggestion
			//get data as json object ->
```

pkg/resolver/searchComplete.go
```diff
@@ -132,14 +138,61 @@ func (s *SearchCompleteResult) searchCompleteResults(ctx context.Context) ([]*st
...
+			var input interface{}
+
+			scanErr := rows.Scan(&input)
 			if scanErr != nil {
 				klog.Error("Error reading searchCompleteResults", scanErr)
 			}
-			srchCompleteOut = append(srchCompleteOut, &prop)
+			switch v := input.(type) {
+			case string:
+				// fmt.Println("prop type: string")
```
jlpadilla, line 152: Delete commented lines below.

pkg/resolver/searchComplete_test.go
```diff
@@ -201,3 +201,60 @@ func Test_SearchCompleteQuery_PropNum(t *testing.T) {
...
+
+	// Mock the database queries.
+	mockRows := newMockRowsWithoutRBAC("../resolver/mocks/mock.json", searchInput, prop1, limit)
+
+	// for _, data := range mockRows.mockData {
+	// 	fmt.Println("Mockdata:", data)
+	// 	for k, val := range data {
+	// 		fmt.Println("Key:", k, "val:", val)
+	// 	}
+	// }
```
jlpadilla, line 225: Delete commented lines.

pkg/resolver/testHelper_test.go
```diff
@@ -96,18 +96,13 @@ func (r *Row) Scan(dest ...interface{}) error {
...
-// ====================================================
-
 //Prop will be the property input for searchComplete
 func newMockRowsWithoutRBAC(mockDataFile string, input *model.SearchInput, prop string, limit int) *MockRows {
 	// Read json file and build mock data
 	bytes, _ := ioutil.ReadFile(mockDataFile)
 	var data map[string]interface{}
 	if err := json.Unmarshal(bytes, &data); err != nil {
-		panic(err)
+		klog.Warning("Error unmarshaling data", err.Error())
```
jlpadilla, line 105: We avoid using panic() in the code because we don't want to kill the pod, but in the tests is ok because this will cause the test to fail later, so it's easier to debug if we panic here.

pkg/resolver/search.go
Anxhela21: Done.

pkg/resolver/testHelper_test.go
Anxhela21: Okay, that makes sense. Will change back to panic. thanks for explaining.

pkg/resolver/search.go
```diff
@@ -334,6 +342,13 @@ func getWhereClauseExpression(prop, operator string, values []string) []exp.Expr
 		}
 	case "=":
 		exps = append(exps, goqu.L(`"data"->>?`, prop).In(values))
+	case "@>":
+		for _, val := range values {
+			exps = append(exps, jsonLookup(goqu.L(`"data"->?`, prop), val))
+		}
+	case "?|":
+		exps = append(exps, jsonAnyKeyArrayLookup(goqu.L(`"data"->?`, prop), values))
```
jlpadilla, line 350: Would this be the same? (Sorry, it may reduce coverage 😢 )
```suggestion
	case "@>":
		for _, val := range values {
			exps = append(exps, goqu.L(`"data"->'?' @> '?'` , prop, val))
		}
	case "?|":
		exps = append(exps, goqu.L(`"data"->'?' ? ?`, prop, "?|", values))
```

**Note:** I need to double check the usage of `?`.  I thought postgres uses `$1, $2, ...`

pkg/resolver/searchComplete.go
```diff
@@ -132,14 +138,59 @@ func (s *SearchCompleteResult) searchCompleteResults(ctx context.Context) ([]*st
...
+			scanErr := rows.Scan(&input)
 			if scanErr != nil {
 				klog.Error("Error reading searchCompleteResults", scanErr)
 			}
-			srchCompleteOut = append(srchCompleteOut, &prop)
+			switch v := input.(type) {
+			case string:
+				prop = v
+				props[v] = struct{}{}
+				// props = append(props, v)
```
jlpadilla, line 154: Delete commented lines. There's about 5 more in this file.

pkg/resolver/search.go
```diff
@@ -310,6 +310,14 @@ func getOperator(values []string) map[string][]string {
...
 func getWhereClauseExpression(prop, operator string, values []string) []exp.Expression {
 	exps := []exp.Expression{}
+	// jsonLookup := func(col exp.Expression, field string) goqu.Expression {
+	// 	return goqu.L("?@>?", col, field)
+	// 	// WHERE ("data"->'label' @> '{"component":"network"}')
+	// }
+	// jsonAnyKeyArrayLookup := func(col exp.Expression, fields []string) goqu.Expression {
+	// 	return goqu.L("???", col, goqu.Literal("?|"), fields)
+	// }
+
```
jlpadilla, line 320: Delete commented lines.
```suggestion
```

pkg/resolver/search.go
```diff
@@ -502,13 +534,34 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
+					for i, val := range values {
+						labels := strings.Split(val, "=")
+						if len(labels) == 2 {
+							cleanedVal[i] = fmt.Sprintf(`{"%s":"%s"}`, labels[0], labels[1])
+						} else if len(labels) == 1 {
+							//// If property is of array type, format it as an array for easy searching
+							cleanedVal[i] = labels[0]
+						} else {
+							klog.Error("Error while decoding label string")
+							cleanedVal[i] = val // <<<REVIEW NOTE: is this the best we can do in this situation?
```
jlpadilla, line 551:
```suggestion
							cleanedVal[i] = val
```

jlpadilla (approved):
/lgtm

- Searching by `label` is working.
- There's a bug when searching by other list properties like container or image.

Merging this PR because it is an incremental improvement and `label` is the most important scenario.  @Anxhela21 and I agreed to fix the bug separately.

jlpadilla:
/unhold
/override ci/prow/sonar-pre-submit

Overriding sonar gate to allow for continuous improvement. Current coverage is 69.4% only 0.6% below target.

PR #89, opened 2022-09-07
rename function

SherinV:

**Related Issue:**  stolostron/backlog#21298

### Description of changes
Rename functions

### DIFF
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -3,6 +3,7 @@ package rbac
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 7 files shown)


pkg/rbac/userData.go
```diff
@@ -111,6 +112,24 @@ func (cache *Cache) GetUserData(ctx context.Context,
 
 }
 
+func (cache *Cache) GetUserData(ctx context.Context) (*UserData, error) {
```
jlpadilla, line 115:
```suggestion
// Get a static copy of the current user data. It will use cached data if valid or refresh if needed.
func (cache *Cache) GetUserData(ctx context.Context) (*UserData, error) {
```

SherinV: Done

jlpadilla (approved):
/lgtm

PR #88, opened 2022-09-06
rbac for messages and related

SherinV:
Signed-off-by: Sherin Varughese <shvarugh@redhat.com>

**Related Issue:**  https://github.com/stolostron/backlog/issues/21298

### Description of changes
Split some code into https://github.com/stolostron/search-v2-api/pull/90 and https://github.com/stolostron/search-v2-api/pull/89

PR #90, opened 2022-09-08
get addon disabled clusters

SherinV:

Related Issue: https://github.com/stolostron/backlog/issues/21298


pkg/resolver/rbacHelper.go
```diff
@@ -83,3 +83,11 @@ func matchManagedCluster(managedClusters []string) exp.BooleanExpression {
 	//managed clusters
 	return goqu.C("cluster").Eq(goqu.Any(pq.Array(managedClusters)))
 }
+
+func getKeysStructMap(structArrayMap map[string]struct{}) []string {
```
SherinV, line 87: This function is removed in the next PR https://github.com/stolostron/search-v2-api/pull/88. It is incorporated into the getKeys() function.

pkg/resolver/messages.go
```diff
@@ -65,27 +72,75 @@ func (s *Message) buildSearchAddonDisabledQuery(ctx context.Context) {
...
 
-func (s *Message) messageResults(ctx context.Context) ([]*model.Message, error) {
-	klog.V(2).Info("Resolving Messages()")
+// If user has access to even one search addon disabled cluster, show the search disabled message
+func (s *Message) checkUserAccessToDisabledClusters(ctx context.Context, disabledClusters *map[string]struct{}) bool {
 
-	rows := s.pool.QueryRow(ctx, s.query)
+	for _, cluster := range getKeysStructMap(*disabledClusters) { //rbac.CacheInst.GetDisabledClusters()) {
+		for _, mc := range s.userData.ManagedClusters {
+			klog.Info("checking user access to cluster: ", cluster)
```
jlpadilla, line 80: Please lower verbosity of log or move before the return.

pkg/resolver/messages.go
```diff
@@ -65,27 +72,75 @@ func (s *Message) buildSearchAddonDisabledQuery(ctx context.Context) {
...
 	if rows != nil {
-		var count int
-		err := rows.Scan(&count)
-		if err != nil {
-			klog.Errorf("Error %s resolving cluster count for query:%s", err.Error(), s.query)
+		for rows.Next() {
+			var srchAddonDisabledCluster string
+			err := rows.Scan(&srchAddonDisabledCluster)
+			if err != nil {
+				klog.Errorf("Error %s resolving cluster count for query: %s", err.Error(), s.query)
```
jlpadilla, line 104: Probably need to `continue` if this error happens.

pkg/resolver/messages.go
```diff
@@ -65,27 +72,75 @@ func (s *Message) buildSearchAddonDisabledQuery(ctx context.Context) {
...
+			}
+			disabledClusters[srchAddonDisabledCluster] = struct{}{}
+
+			//Since cache is not valid, update shared cache with disabled clusters result
+			rbac.CacheInst.SetDisabledClusters(disabledClusters, nil)
```
jlpadilla, line 109: I think this needs to be moved after the for loop.

pkg/resolver/messages.go
```diff
@@ -65,27 +72,75 @@ func (s *Message) buildSearchAddonDisabledQuery(ctx context.Context) {
...
+	} else {
+		klog.V(5).Info("DisabledClusters Cache not valid - running query to get search addon disabled clusters")
+
+		if disabledClusters, err := s.findSrchAddonDisabledClusters(ctx); err != nil {
+			klog.Error("Error retrieving Search Addon disabled clusters: ", err)
+			return []*model.Message{}, err
+		} else {
+			returnDisabledMessage = s.checkUserAccessToDisabledClusters(ctx, disabledClusters)
 		}
 	}
```
jlpadilla, line 130: This logic should be part of the cache. 
From here we should call something like `rbac.cacheInst.getDisabledClusters()`
Then inside that function we'll check if we have the data in the cache or if we need to request it.

pkg/rbac/sharedData.go
```diff
@@ -210,3 +221,24 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context)
...
+	} else {
+		klog.V(5).Info("DisabledClusters Cache not valid") // - running query to get search addon disabled clusters")
+		cache.SetDisabledClusters(map[string]struct{}{}, fmt.Errorf("addon disabled clusters cache is invalid"))
+	}
+	cache.shared.dcLock.Lock()
+	defer cache.shared.dcLock.Unlock()
+	return &cache.shared.disabledClusters, cache.shared.dcErr
+}
+
+func (cache *Cache) SetDisabledClusters(disabledClusters map[string]struct{}, err error) {
```
jlpadilla, line 238: Keep this function private (lowercase).  Data should only get set within this package.
```suggestion
func (cache *Cache) setDisabledClusters(disabledClusters map[string]struct{}, err error) {
```

jlpadilla: This change might trigger some refactoring because I see `SetDisabledClusters()` is getting called from messages.go

Ideally `messages.go` would just request the disabled clusters and it should be up to the cache logic to return existing cached data or trigger a request to refresh the data.

pkg/resolver/messages.go
```diff
@@ -65,18 +72,78 @@ func (s *Message) buildSearchAddonDisabledQuery(ctx context.Context) {
...
+			rbac.CacheInst.SetDisabledClusters(map[string]struct{}{}, err)
+			return &map[string]struct{}{}, err
+		} else {
+			rbac.CacheInst.SetDisabledClusters(*disabledClustersFromQuery, nil)
+			return disabledClustersFromQuery, err
+		}
+	} else {
+		klog.Info("Disabled clusters cache valid. Returning results")
+	}
+	//cache is valid
```
jlpadilla, line 129: I'd expect the logic in this block to be part of the cache.

pkg/rbac/sharedData.go
```diff
@@ -81,6 +87,11 @@ func (cache *Cache) PopulateSharedCache(ctx context.Context) error {
 
 }
 
+func (cache *Cache) SharedCacheDisabledClustersValid() bool {
```
jlpadilla, line 90: Make private (lowercase)
```suggestion
func (cache *Cache) sharedCacheDisabledClustersValid() bool {
```

pkg/rbac/sharedData.go
```diff
@@ -210,3 +221,133 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context)
 	return shared.mcErr
 
 }
+
+func (cache *Cache) GetDisabledClusters(ctx context.Context) (*map[string]struct{}, error) {
```
jlpadilla, line 225:
```suggestion
// Returns a map of managed clusters for which the search add-on has been disabled.
func (cache *Cache) GetDisabledClusters(ctx context.Context) (*map[string]struct{}, error) {
```

pkg/rbac/sharedData.go
```diff
@@ -210,3 +221,133 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context)
...
+	userData, userDataErr := cache.GetUserData(ctx)
+	if userDataErr != nil {
+		return nil, userDataErr
+	}
+
+	if cache.SharedCacheDisabledClustersValid() {
+		klog.V(5).Info("Search Addon DisabledClusters Cache valid")
+	} else {
+		klog.V(5).Info("DisabledClusters Cache not valid") // - running query to get search addon disabled clusters")
+		//run query and get disabled clusters
```
jlpadilla, line 235: Need to lock before running the query. otherwise we could execute the same query multiple times if we get concurrent requests.

SherinV:
Getting this error in unit-test while running in prow.

> Error resolving TokenReview from Kube API.tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:ci-op-******-default" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope

pkg/rbac/sharedData.go
```diff
@@ -210,3 +221,133 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context)
...
+			cache.setDisabledClusters(*disabledClustersFromQuery, nil)
+		}
+	}
+
+	//check if user has access to disabled clusters
+	if userHasAccessToDisabledClusters(&cache.shared.disabledClusters, userData.ManagedClusters) {
+		klog.V(5).Info("user has access to Search Addon disabled clusters ")
+		cache.shared.dcLock.Lock()
+		defer cache.shared.dcLock.Unlock()
+		return &cache.shared.disabledClusters, cache.shared.dcErr
```
jlpadilla, line 250: When the user has access to 1 disabled cluster, we will return the full list of disabled clusters, which may include clusters not visible to the user.
Need to either:
- Return the correct list OR
- Update the function to return a boolean instead. (This might be enough for the messages scenario.)

pkg/rbac/sharedData.go
```diff
@@ -210,3 +221,133 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context)
...
+func (cache *Cache) GetDisabledClusters(ctx context.Context) (*map[string]struct{}, error) {
+	userData, userDataErr := cache.GetUserData(ctx)
+	if userDataErr != nil {
+		return nil, userDataErr
+	}
+
+	if cache.SharedCacheDisabledClustersValid() {
+		klog.V(5).Info("Search Addon DisabledClusters Cache valid")
+	} else {
+		klog.V(5).Info("DisabledClusters Cache not valid") // - running query to get search addon disabled clusters")
```
jlpadilla, line 234: This is simpler and reduces log verbosity (no log means we used cached data).
```suggestion
	if !cache.SharedCacheDisabledClustersValid() {
		klog.V(5).Info("DisabledClusters cache empty or expired. Querying database.")
```

jlpadilla:
@SherinV This means that there's a test using the real kubernetes client instead of a mock kube client.

> Getting this error in unit-test while running in prow.
> 
> > Error resolving TokenReview from Kube API.tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:ci-op-******-default" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope

pkg/rbac/sharedData.go
SherinV: Done, thanks.

pkg/rbac/sharedData.go
```diff
@@ -210,3 +221,137 @@ func (shared *SharedData) GetManagedClusters(cache *Cache, ctx context.Context)
...
+		return &userAccessClusters, cache.shared.dcErr
+
+	} else {
+		klog.V(5).Info("user does not have access to Search Addon disabled clusters ")
+		return &map[string]struct{}{}, nil
+	}
+
+}
+
+func userHasAccessToDisabledClusters(disabledClusters *map[string]struct{}, userClusters []string) (bool,
```
jlpadilla, line 262: - Can simplify removing the returned bool. It's the same as `len(userAccessDisabledClusters) > 0`
- Maybe rename to something like `disabledClustersForUser()

pkg/rbac/cache.go
```diff
@@ -25,7 +25,7 @@ type Cache struct {
 	// Defining these here allow the tests to replace with a mock client.
 	authnClient   authnv1.AuthenticationV1Interface
 	corev1Client  corev1.CoreV1Interface
-	pool          pgxpoolmock.PgxPool // Database client
+	Pool          pgxpoolmock.PgxPool // Database client
```
jlpadilla, line 28: Do we still need this to be public? (excluding tests for now.)

jlpadilla:
We are having challenges mocking the cache to tests messages.go, so I suggested to Sherin we merge this and work on that problem separately. That will help us to isolate the test problem and also continue to make progress.

jlpadilla (approved):
/lgtm
/hold

PR #91, opened 2022-09-16
Cache and handle all property types

Anxhela21:

**Related Issue:**  https://issues.redhat.com/browse/ACM-1742

### Description of changes
- Implement search queries using `label` and other properties with list values.


pkg/config/config.go
```diff
@@ -16,20 +16,21 @@ var Cfg = new()
...
-	DBUser         string
-	HttpPort       int
-	PlaygroundMode bool // Enable the GraphQL Playground client.
-	QueryLimit     int  // The default LIMIT to use on queries. Client can override.
-	RelationLevel  int  // The number of levels/hops for finding relationships for a particular resource
+	API_SERVER_URL    string // address for Kubernetes API Server
+	AuthCacheTTL      int    // Time-to-live (milliseconds) of Authentication (TokenReview) cache.
+	SharedCacheTTL    int    // Time-to-live (milliseconds) of common resources (shared across users) cache.
+	UserCacheTTL      int    // Time-to-live (milliseconds) of namespaced resources (specifc to users) cache.
+	PropTypesCacheTTL int    // Time-to-live (milliseconds) of property datatypes cache in shared cache.
```
jlpadilla, line 23: I'd reuse `SharedCacheTTL` to keep it simple.
This is better, but I'm afraid that too many different TTL will create confusion.

pkg/rbac/sharedData.go
```diff
@@ -51,16 +55,86 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+
+	var key string
+	var value string
+	resourceTypeMap := make(map[string]string)
+	for rows.Next() {
+		err = rows.Scan(&key, &value)
+
+		if err != nil {
+			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), query)
+		}
```
jlpadilla, line 94: If there's an error we can't process this row, so need to add `continue`.
```suggestion
		if err != nil {
			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), query)
			continue
		}
```

pkg/rbac/sharedData.go
```diff
@@ -51,16 +55,86 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	//select from these datasets
+	ds := goqu.From(schemaTable, jsb)
+	//select statement with orderby and distinct clause
+	selectDs = ds.Select(goqu.L("key"), goqu.L("jsonb_typeof(?)",
+		goqu.C("value")).As("datatype")).Distinct().Order(goqu.L("datatype").Asc())
+
+	query, params, err := selectDs.ToSQL()
+
+	klog.V(5).Info("Query for property datatypes: [%s] ", query)
+	rows, err := CacheInst.pool.Query(context.Background(), query, params...)
```
jlpadilla, line 79: This function should get a context as an argument instead of creating a new one here.

pkg/rbac/sharedData.go
```diff
@@ -51,16 +55,86 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	jsb := goqu.L("jsonb_each(?)", goqu.C("data"))
+	//select from these datasets
+	ds := goqu.From(schemaTable, jsb)
+	//select statement with orderby and distinct clause
+	selectDs = ds.Select(goqu.L("key"), goqu.L("jsonb_typeof(?)",
+		goqu.C("value")).As("datatype")).Distinct().Order(goqu.L("datatype").Asc())
+
+	query, params, err := selectDs.ToSQL()
+
+	klog.V(5).Info("Query for property datatypes: [%s] ", query)
```
jlpadilla, line 78:
```suggestion
	klog.V(5).Infof("Query for property datatypes: [%s] ", query)
```

pkg/rbac/sharedData.go
```diff
@@ -51,16 +55,86 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
 	Resource: "managedclusters",
 }
 
-func (cache *Cache) PopulateSharedCache(ctx context.Context) error {
+func (shared *SharedData) GetPropertyTypes() (map[string]string, error) {
```
jlpadilla, line 58: I think this can be private.
```suggestion
func (shared *SharedData) getPropertyTypes() (map[string]string, error) {
```

pkg/config/config.go
Anxhela21: sounds good.

pkg/rbac/sharedData.go
Anxhela21: Done.

pkg/rbac/sharedData.go
Anxhela21: Yes, meant to change that back to private.

Anxhela21:
Noticing some inconsistencies from the sonarcloud report and the code currently merged. Some of the code that was merged from main in this PR does **not** show up in the sonarcloud report.

pkg/rbac/sharedData.go
```diff
@@ -51,36 +59,112 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+		klog.Errorf("Error resolving query [%s] with args [%+v]. Error: [%+v]", query, err)
+		return nil, err
+	}
+	defer rows.Close()
+
+	var key string
+	var value string
+	resourceTypeMap := make(map[string]string)
+	for rows.Next() {
+		err = rows.Scan(&key, &value)
```
jlpadilla, line 94: Declare these inside the loop.
FYI, this is a recent bug fix for a similar problem. https://github.com/stolostron/search-v2-api/pull/98
```suggestion
	resourceTypeMap := make(map[string]string)
	for rows.Next() {
		var key, value string
		err = rows.Scan(&key, &value)
```

pkg/rbac/sharedData.go
```diff
@@ -51,36 +59,112 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+
+		if err != nil {
+			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), query)
+			continue
+		}
+		resourceTypeMap[key] = value
+
+	}
+	defer rows.Close()
```
jlpadilla, line 103: Delete this line because this was done above, around line 88.
```suggestion
```

pkg/rbac/sharedData.go
```diff
@@ -51,36 +59,112 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
-		return nil
+		return &cache.shared, nil
 	} else { //get data and cache
 
 		var error error
+		_, err := cache.shared.getPropertyTypes(cache, ctx)
+		if cache.shared.propTypeErr == nil {
+			klog.V(6).Info("Successfully retrieved cluster scoped resources!")
+		} else {
+			error = err
```
jlpadilla, line 142: If there's more than 1 error in the logic below, only the last error will be returned.
I'd change this to log only if there's an error instead of logging on success. Logging errors is more helpful for debugging.

Anxhela21: That's a good point. Will log on error instead. This entire function needs to be changed so that these are running in parallel (in another PR)

pkg/rbac/sharedData.go
```diff
@@ -38,6 +39,13 @@ type SharedData struct {
...
 	nsLock      sync.Mutex // Locks the namespaces array while updating it.
 	nsUpdatedAt time.Time  // Time when namespaces data was last updated.
+
+	propTypeErr  error
+	propTypeTime time.Time
+}
+
+type PropTypesCache struct {
+	propType map[string]string //list of all properties and thier relative data types among all users.
 }
```
jlpadilla, line 49: - Why is this needed instead of declaring `propType map[string]string` directly into sharedData?
- If we are declaring a new struct, then I'd combine the error and type like this.
    ```go
	type PropTypesCache struct {
		propType map[string]string //list of all properties and thier relative data types among all users.
		propTypeErr  error
		propTypeTime time.Time
	}
    ```

While both alternatives work, try to keep consistency with the existing code. Or consider refactoring existing code to keep consistency.

Anxhela21: I guess we don't need to keep it separate from main cache. I was trying to be consistent with userDataCache and how we store the resources in a separate struct for easy access.

pkg/rbac/sharedData.go
```diff
@@ -51,36 +55,108 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	}
+	//cache query:
+	shared.propTypes = resourceTypeMap
+	shared.propTypeErr = err
+	shared.propTypeTime = time.Now()
+
+	return resourceTypeMap, err
+}
+
+func (cache *Cache) GetSharedData(ctx context.Context) (map[string]string, error) {
```
jlpadilla, line 105: Rename this function. From `GetSharedData()` I assumed it would return all data in the shared cache, but it only returns the propertyType.

pkg/rbac/sharedData.go
```diff
@@ -51,36 +55,108 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	//store only the PropTypeCache to use in outside of rbac module
+	SharedDataAccess := sharedData.GetPropertyTypeCache()
+
+	propTypesMap := SharedDataAccess
+	return propTypesMap, nil
+}
+
+func (shared *SharedData) GetPropertyTypeCache() map[string]string {
+	return shared.propTypes
+}
```
jlpadilla, line 120: Do we need this function? I only see 1 reference, which could be refactored without the additional function.
image
```suggestion
	//store only the PropTypeCache to use in outside of rbac module
	propTypesMap := shared.propTypes
	return propTypesMap, nil
}
```

Anxhela21: Done.

pkg/resolver/search.go
```diff
@@ -526,39 +548,44 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
 
+						for i, val := range values {
+							if dataType == "object" { //map[string]interface{}
+								labels := strings.Split(val, "=")
+								cleanedVal[i] = fmt.Sprintf(`{"%s":"%s"}`, labels[0], labels[1])
+							} else if dataType == "array" { //[]string
+								cleanedVal[i] = fmt.Sprintf(`["%s"]`, val)
+
+							} else {
+								klog.Error("Error while decoding label string")
```
jlpadilla, line 571: Add more information to help debug.
```suggestion
								klog.Errorf("Error while decoding label string with dataType %s", dataType)
```

pkg/resolver/searchComplete.go
```diff
@@ -44,13 +45,21 @@ func SearchComplete(ctx context.Context, property string, srchInput *model.Searc
 	if userDataErr != nil {
 		return []*string{}, userDataErr
 	}
+
+	//check that shared cache has resource datatypes:
+	propTypesCache, err := rbac.GetCache().GetPropertyTypes(ctx)
+	if err != nil {
+		klog.Warningf("Error creating datatype map with err: [%s] ", err)
```
jlpadilla, line 52: What happens if we can't get the propertyTypes? Should we "fall back" to only strings?

pkg/resolver/searchComplete_test.go
```diff
@@ -238,3 +236,53 @@ func Test_SearchCompleteWithLabel_Query(t *testing.T) {
 	// Verify response
 	AssertStringArrayEqual(t, result, expectedProps, "Error in Test_SearchCompleteWithLabel_Query")
 }
+
```
jlpadilla, line 239: Do we have a test that covers `label`?

Anxhela21: yes, label is covered in `Test_SearchCompleteWithLabel_Query()`

pkg/rbac/sharedData.go
```diff
@@ -95,7 +166,8 @@ func sharedCacheValid(shared *SharedData) bool {
 
 	if (time.Now().Before(shared.csUpdatedAt.Add(time.Duration(config.Cfg.SharedCacheTTL) * time.Millisecond))) &&
 		(time.Now().Before(shared.nsUpdatedAt.Add(time.Duration(config.Cfg.SharedCacheTTL) * time.Millisecond))) &&
-		(time.Now().Before(shared.mcUpdatedAt.Add(time.Duration(config.Cfg.SharedCacheTTL) * time.Millisecond))) {
+		(time.Now().Before(shared.mcUpdatedAt.Add(time.Duration(config.Cfg.SharedCacheTTL) * time.Millisecond))) &&
+		(time.Now().Before(shared.propTypeTime.Add(time.Duration(config.Cfg.SharedCacheTTL) * time.Millisecond))) {
```
SherinV, line 170: Do we need to invalidate the cache based on property types updated time? I hope refreshing property types if a  property is not found in there would suffice?

pkg/rbac/sharedData.go
```diff
@@ -51,36 +55,103 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
 		if err == nil {
-			klog.V(6).Info("Successfully retrieved managed clusters!")
+			klog.Errorf("Error retrieving managed clusters. Error: [%+v]", err)
 		} else {
-			error = err
+			klog.V(6).Info("Successfully retrieved managed clusters!")
 		}
-		return error
+
+		return &cache.shared, error
```
SherinV, line 154: Do shared cache need to be returned here? It is part of the cache and can be accessed from there.

pkg/resolver/search.go
```diff
@@ -508,39 +529,44 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
-						if len(labels) == 2 {
-							cleanedVal[i] = fmt.Sprintf(`{"%s":"%s"}`, labels[0], labels[1])
-						} else if len(labels) == 1 {
-							//// If property is of array type, format it as an array for easy searching
-							cleanedVal[i] = labels[0]
-						} else {
-							klog.Error("Error while decoding label string")
-							cleanedVal[i] = val
-						}
+				for key, val := range propTypeMap {
```
SherinV, line 537: Lookup the map instead of looping through it.

> dataType, ok := propTypeMap[filter.Property]

pkg/resolver/search.go
```diff
@@ -508,39 +529,44 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
+					if key == filter.Property { //check if property exists in dataTypeMap to get datatype
```
SherinV, line 538: What happens if `filter.Property` is not present in propTypeMap? This was where I was hoping to repopulate shared cache property types (querying db again).

pkg/rbac/sharedData.go
jlpadilla: I like @SherinV 's approach of "appending to the cache" whenever we encounter a new property.  Currently we don't have any scenario where the property type changes after we have discovered it.
However, this could be a significant change. We can try on a new branch but don't discard the current implementation.

Anxhela21: This is a good point. Should we continue with current implementation and add that as an enhancement?

pkg/rbac/sharedData.go
Anxhela21: we can call on cache.

SherinV: Ok.  remove the &cache.shared from the return.

pkg/resolver/search.go
```diff
@@ -167,8 +177,13 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
 
 	//WHERE CLAUSE
 	if s.input != nil && (len(s.input.Filters) > 0 || (s.input.Keywords != nil && len(s.input.Keywords) > 0)) {
```
SherinV, line 179: Isn't this the usual flow? Why is the RBAC clause moved to else clause?

pkg/resolver/search.go
```diff
@@ -167,8 +177,13 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
...
-		whereDs = WhereClauseFilter(s.input)
-		sql, _, err := selectDs.Where(whereDs...).ToSQL()
+
+		//we get properties here as well as their datatypes:
+		whereDs, typeFilter = WhereClauseFilter(s.input, s.propTypes)
+		klog.V(7).Info("Property Datatype", typeFilter)
+
+	} else {
```
SherinV, line 185: What does the else clause mean here?

SherinV: This is why the RBAC clauses are excluded in the tests.

Anxhela21: Ah, okay. Not sure why it got excluded. Thanks for taking a look!

pkg/rbac/sharedData.go
```diff
@@ -51,36 +55,107 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+		klog.Error("Error populating shared data cache: ", err)
+		return nil, err
+	}
+	//store only the PropTypeCache to use in outside of rbac module
+	propTypesMap := sharedData.propTypes
+
+	return propTypesMap, nil
+}
+
+func (cache *Cache) PopulateSharedCache(ctx context.Context) (*SharedData, error) {
```
SherinV, line 121:
```suggestion
func (cache *Cache) PopulateSharedCache(ctx context.Context) (error) {
```

pkg/rbac/sharedData.go
```diff
@@ -51,36 +55,107 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
 		if err == nil {
-			klog.V(6).Info("Successfully retrieved managed clusters!")
+			klog.Errorf("Error retrieving managed clusters. Error: [%+v]", err)
 		} else {
-			error = err
+			klog.V(6).Info("Successfully retrieved managed clusters!")
 		}
-		return error
+
+		return &cache.shared, error
```
SherinV, line 158:
```suggestion
		return  error
```

pkg/resolver/searchComplete_test.go
SherinV: Can you add a test case for general search too? This one is SearchComplete.

SherinV: I see an issue with image `quay.io/stolostron/search-v2-api:2.7.0-PR91-4886e6c66e49108ff1bfb05e1ca4bfb1cb8e22c4` on my cluster
Step 1: 
Search with label.
image

Step 2:
Add `cluster` filter
image

SherinV:
Another issue with an array property - `port`.

Step 1: works fine. Notice the port in the result.

image

Step 2: Search for port `4194/TCP`
image

Currently searching for port works if the array has only one element.


Same with another array property: `role`

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,68 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	//select from these datasets
+	ds := goqu.From(schemaTable, jsb)
+
+	//select statement with orderby and distinct clause
+	selectDs = ds.Select(goqu.L("key"), goqu.L("jsonb_typeof(?)",
+		goqu.C("value")).As("datatype")).Distinct()
+
+	query, params, sqlerr := selectDs.ToSQL()
+	if sqlerr != nil {
+		klog.Errorf("Error building Search query: %s", sqlerr.Error())
```
SherinV, line 78: We cannot proceed if query building has an error.

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,68 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	}
+
+	klog.V(5).Infof("Query for property datatypes: [%s] ", query)
+	rows, err := cache.pool.Query(ctx, query, params...)
+	if err != nil {
+		klog.Errorf("Error resolving query [%s] with args [%+v]. Error: [%+v]", query, err)
+		return nil, err
```
SherinV, line 85: Return an empty map or check for nil before you access the map later in the code.

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,68 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	}
+	defer rows.Close()
+	resourceTypeMap := make(map[string]string)
```
SherinV, line 88: Move this to the top and return that instead of nil to avoid nil pointer error later.

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,68 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	for rows.Next() {
+		var key, value string
+		err = rows.Scan(&key, &value)
+
+		if err != nil {
+			klog.Errorf("Error %s retrieving rows for query:%s", err.Error(), query)
```
SherinV, line 94:
```suggestion
			klog.Errorf("Error %s scanning value for getPropertyTypes query:%s", err.Error(), query)
```

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,68 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	//cache query:
+	shared.propTypes = resourceTypeMap
+	shared.propTypeErr = err
+	shared.propTypeTime = time.Now()
+
+	return resourceTypeMap, err
+}
+
+func (cache *Cache) GetPropertyTypes(ctx context.Context) (map[string]string, error) {
+	err := cache.PopulateSharedCache(ctx)
```
SherinV, line 109: We don't have to populate all the other shared caches when only prop type is needed. Because the proptype can be called at a different frequency than the other ones - depends on if the property itself is present in the map.

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,68 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	if err != nil {
+		klog.Error("Error populating shared data cache: ", err)
+		return nil, err
```
SherinV, line 112: return an empty map if you are not doing the nil check later.

pkg/rbac/sharedData.go
```diff
@@ -59,27 +125,35 @@ func (cache *Cache) PopulateSharedCache(ctx context.Context) error {
...
+		}
 		// get all cluster-scoped resources and cache in shared.csResources
-		err := cache.shared.GetClusterScopedResources(cache, ctx)
-		if err == nil {
-			klog.V(6).Info("Successfully retrieved cluster scoped resources!")
+		err = cache.shared.GetClusterScopedResources(cache, ctx)
+		if err != nil {
+			klog.Errorf("Error retrieving cluster scoped resources. Error: [%+v]", err)
 		} else {
-			error = err
```
SherinV, line 67: Don't remove this or do return the err variable.

pkg/rbac/sharedData.go
```diff
@@ -59,27 +125,35 @@ func (cache *Cache) PopulateSharedCache(ctx context.Context) error {
...
 		err = cache.shared.GetManagedClusters(cache, ctx)
 		if err == nil {
-			klog.V(6).Info("Successfully retrieved managed clusters!")
+			klog.Errorf("Error retrieving managed clusters. Error: [%+v]", err)
 		} else {
-			error = err
+			klog.V(6).Info("Successfully retrieved managed clusters!")
 		}
+
 		return error
```
SherinV, line 157: Currently this is empty, even in case of error.

pkg/resolver/mocks/mock.json
```diff
@@ -31,7 +31,6 @@
             "desired":3,
             "kind":"ReplicaSet",
             "label":{"pod-template-hash":"5f5575c669"},
-            "container": ["acm-agent", "acm-agent-2"],
```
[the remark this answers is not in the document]
SherinV: Testing for `acm-agent` should bring back this Replicaset.

pkg/resolver/search.go
```diff
@@ -397,20 +421,15 @@ func getOperatorAndNumDateFilter(filter string, values []string) map[string][]st
...
 
 				} else {
-					if _, ok := arrayProperties[filter]; ok {
-						klog.V(7).Info("filter ", filter, " is present in arrayProperties. Using operator ?|.")
-						operator = "?|"
-					} else {
-						klog.V(7).Info("filter is neither label nor in arrayProperties: ", filter)
-						operator = ""
-					}
+					klog.V(7).Info("filter is neither label nor in arrayProperties: ", filter)
```
SherinV, line 429:
```suggestion

```

pkg/resolver/search.go
```diff
@@ -508,49 +533,66 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
-				for _, operator := range keys {
-					operatorWhereDs = append(operatorWhereDs,
-						getWhereClauseExpression(filter.Property, operator, opDateValueMap[operator])...)
+					whereDs = append(whereDs, goqu.Or(operatorWhereDs...)) //Join all the clauses with OR
 				}
-				whereDs = append(whereDs, goqu.Or(operatorWhereDs...)) //Join all the clauses with OR
-
 			} else {
 				klog.Warningf("Ignoring filter [%s] because it has no values", filter.Property)
+				// }
```
SherinV, line 567:
```suggestion

```

pkg/resolver/search.go
```diff
@@ -508,49 +533,66 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
 			}
 		}
 	}
 
-	return whereDs
+	return whereDs, dataTypeFromMap
+}
+
+func decodePropertyTypes(values []string, dataTypeFromMap string) ([]string, string) {
+	var dataType string
```
SherinV, line 576:
```suggestion

```

pkg/resolver/search.go
```diff
@@ -508,49 +533,66 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
+
+	for i, val := range values {
+		if dataType == "object" {
+			labels := strings.Split(val, "=")
+			cleanedVal[i] = fmt.Sprintf(`{"%s":"%s"}`, labels[0], labels[1])
+		} else if dataType == "array" {
+			cleanedVal[i] = fmt.Sprintf(`["%s"]`, val)
+
+		} else {
+			klog.Errorf("Error while decoding label string with dataType %s", dataType)
```
SherinV, line 588:
```suggestion

```

pkg/resolver/search.go
```diff
@@ -508,49 +533,66 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
-						} else if len(labels) == 1 {
-							//// If property is of array type, format it as an array for easy searching
-							cleanedVal[i] = labels[0]
-						} else {
-							klog.Error("Error while decoding label string")
-							cleanedVal[i] = val
-						}
+				if dataTypeInMap, ok := propTypeMapFromCache[filter.Property]; !ok { //check if value exists/if value doesn't exist
+					klog.Warningf("Property type for [%s] doesn't exist property type map. Refreshing property type cache", filter.Property)
+					propTypeMap, _ := getPropertyTypeCache(ctx) //call database cache again
```
SherinV, line 543: Do we want to return the new updated map so that SearchResult resolver has the new map?

SherinV, line 543: What happens if there is an error here?

pkg/resolver/searchComplete.go
```diff
@@ -69,9 +78,10 @@ func (s *SearchCompleteResult) searchCompleteQuery(ctx context.Context) {
 	schemaTable := goqu.S("search").Table("resources")
 	ds := goqu.From(schemaTable)
 	if s.property != "" {
+
 		//WHERE CLAUSE
 		if s.input != nil && len(s.input.Filters) > 0 {
-			whereDs = WhereClauseFilter(s.input)
+			whereDs, _ = WhereClauseFilter(ctx, s.input, s.propTypes)
```
SherinV, line 84: If there is an updated map, search result can use that.

pkg/resolver/testHelper_test.go
```diff
@@ -423,3 +427,18 @@ func AssertStringArrayEqual(t *testing.T, result, expected []*string, message st
 		}
 	}
 }
+
+func AssertStringArrayListEqual(t *testing.T, result []*string, expected []*string, message string) {
```
SherinV, line 431: same as  AssertStringArrayEqual

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,72 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	}
+	//cache query:
+	shared.propTypes = resourceTypeMap
+	shared.propTypeErr = err
+	shared.propTypeTime = time.Now()
+
+	return resourceTypeMap, err
+}
+
+func (cache *Cache) GetPropertyTypes(ctx context.Context) (map[string]string, error) {
```
jlpadilla, line 109: I expect this function to return cached data if it exists and valid.
Otherwise it runs the query to refresh the data.

pkg/resolver/search.go
```diff
@@ -157,38 +162,42 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
...
-	if count {
-		selectDs = ds.Select(goqu.COUNT("uid"))
-	} else if uid {
-		selectDs = ds.Select("uid")
-	} else {
-		selectDs = ds.SelectDistinct("uid", "cluster", "data")
-	}
-
-	//WHERE CLAUSE
 	if s.input != nil && (len(s.input.Filters) > 0 || (s.input.Keywords != nil && len(s.input.Keywords) > 0)) {
```
jlpadilla, line 165: We need an `else` to return an error if this validation isn't met.

pkg/resolver/search.go
```diff
@@ -265,321 +276,69 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
-func pointerToStringArray(pointerArray []*string) []string {
-
-	values := make([]string, len(pointerArray))
-	for i, val := range pointerArray {
-		values[i] = *val
-	}
-	return values
-}
+func WhereClauseFilter(ctx context.Context, input *model.SearchInput,
+	propTypeMap map[string]string) ([]exp.Expression, map[string]string, error) {
```
jlpadilla, line 280: Is this better?
```suggestion
func (s *SearchResult) WhereClauseFilter(ctx context.Context, input *model.SearchInput) ([]exp.Expression, error) {
```

jlpadilla: Also `context` is in the SearchResult struct.
```suggestion
func WhereClauseFilter(input *model.SearchInput,
	propTypeMap map[string]string) ([]exp.Expression, map[string]string, error) {
	
	// access with 
	// s.propTypeMap
	// s.context
```

pkg/resolver/search.go
```diff
@@ -265,321 +276,69 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
-						} else {
-							klog.Error("Error while decoding label string")
-							cleanedVal[i] = val
+				values := PointerToStringArray(filter.Values)
+
+				if len(propTypeMap) > 0 {
+					if dataTypeInMap, ok := propTypeMap[filter.Property]; !ok { //check if value exists/if value doesn't exist
+						klog.Warningf("Property type for [%s] doesn't exist in cache. Refreshing property type cache",
+							filter.Property)
+						propTypeMapNew, err := GetPropertyTypeCache(ctx) //call database cache again
```
jlpadilla, line 305: We need to pass an argument to force the cache to refresh.

pkg/resolver/search.go
```diff
@@ -265,321 +276,69 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
-				for _, operator := range keys {
-					operatorWhereDs = append(operatorWhereDs,
-						getWhereClauseExpression(filter.Property, operator, opDateValueMap[operator])...)
+						whereDs = append(whereDs, goqu.Or(operatorWhereDs...)) //Join all the clauses with OR
+						fmt.Println(whereDs)
+					}
+				} else {
+					//if map is empty don't return anything
+					klog.Error("Error with property type list is empty.")
+					return nil, nil, err
```
jlpadilla, line 334: If there's any error getting the property type, we should fall back to use the existing logic. This reduces the impact to users in case there's a problem because only a few properties are lists.

pkg/rbac/sharedData.go
Anxhela21: Done.

pkg/resolver/search.go
Anxhela21: Done.

pkg/resolver/search.go
Anxhela21: I'm not sure this would work, because we use this where clause for search complete too.

pkg/resolver/search.go
Anxhela21: Done.

pkg/resolver/search.go
```diff
@@ -265,321 +276,69 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
+						}
 
-				//Sort map according to keys - This is for the ease/stability of tests when there are multiple operators
-				keys := getKeys(opDateValueMap)
-				var operatorWhereDs []exp.Expression //store all the clauses for this filter together
-				for _, operator := range keys {
-					operatorWhereDs = append(operatorWhereDs,
-						getWhereClauseExpression(filter.Property, operator, opDateValueMap[operator])...)
+						whereDs = append(whereDs, goqu.Or(operatorWhereDs...)) //Join all the clauses with OR
+						fmt.Println(whereDs)
```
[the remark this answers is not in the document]
Anxhela21: meant to be deleted. Done.

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,82 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	shared.propTypes = propTypeMap
+	shared.propTypeErr = err
+	shared.propTypeTime = time.Now()
+
+	return propTypeMap, err
+}
+
+func (cache *Cache) GetPropertyTypes(ctx context.Context, refresh bool) (map[string]string, error) {
+
+	// propTypesMap := make(map[string]string)
```
jlpadilla, line 111: delete
```suggestion
```

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,82 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+		err = rows.Scan(&key, &value)
+
+		if err != nil {
+			klog.Errorf("Error %s scanning value for getPropertyTypes:%s", err.Error(), query)
+			continue
+		}
+		propTypeMap[key] = value
+
+	}
+	//cache query:
```
jlpadilla, line 101:
```suggestion
	// cache results
```

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,82 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	if len(cache.shared.propTypes) > 0 && cache.shared.propTypeErr == nil && !refresh {
+		klog.V(6).Info("Using property types from cache.")
+
+		propTypesMap := cache.shared.propTypes
+		return propTypesMap, nil
+
+	} else {
+		//run query to refresh data
+		propTypes, err := cache.shared.getPropertyTypes(cache, ctx)
+		if err != nil {
```
jlpadilla, line 122: Move this log to reduce how much data we log.
```suggestion
	if len(cache.shared.propTypes) > 0 && cache.shared.propTypeErr == nil && !refresh {
		propTypesMap := cache.shared.propTypes
		return propTypesMap, nil
	} else {
		// run query to refresh data
		klog.V(6).Info("Getting property types from database.")
		propTypes, err := cache.shared.getPropertyTypes(cache, ctx)
		if err != nil {
```

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,82 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+			klog.Errorf("Error retrieving property resources. Error: [%+v]", err)
+			return map[string]string{}, err
+		} else {
+			klog.V(6).Info("Successfully retrieved property types!")
+			// propTypesMap = propTypes
+
```
jlpadilla, line 128: Delete
```suggestion
```

pkg/rbac/sharedData.go
```diff
@@ -130,10 +214,11 @@ func (shared *SharedData) GetClusterScopedResources(cache *Cache, ctx context.Co
...
 
-	rows, queryerr := cache.pool.Query(ctx, query)
-	if queryerr != nil {
-		klog.Errorf("Error resolving query [%s]. Error: [%+v]", query, queryerr.Error())
-		shared.csErr = queryerr
+	// var queryerr error
+	rows, err := cache.pool.Query(ctx, query)
+	if err != nil {
+		klog.Errorf("Error resolving query [%s]. Error: [%+v]", query, err.Error())
+		shared.csErr = err
```
jlpadilla, line 221: Delete comment and update log message.
```suggestion
	rows, err := cache.pool.Query(ctx, query)
	if err != nil {
		klog.Errorf("Error resolving cluster scoped resources. Query [%s]. Error: [%+v]", query, err.Error())
		shared.csErr = err
```

pkg/resolver/search.go
jlpadilla: ok, thanks for clarifying.

pkg/resolver/search.go
```diff
@@ -265,321 +278,72 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
...
-							klog.Error("Error while decoding label string")
-							cleanedVal[i] = val
+				values := PointerToStringArray(filter.Values)
+
+				if len(propTypeMap) > 0 {
+					if dataTypeInMap, ok := propTypeMap[filter.Property]; !ok { //check if value exists/if value doesn't exist
+						klog.Warningf("Property type for [%s] doesn't exist in cache. Refreshing property type cache",
+							filter.Property)
+						refresh := true
+						propTypeMapNew, err := GetPropertyTypeCache(ctx, refresh) //call database cache again
```
jlpadilla, line 309: No need to declare a new variable.
```suggestion
						propTypeMapNew, err := GetPropertyTypeCache(ctx, true) // call database cache again
```

pkg/resolver/searchComplete.go
```diff
@@ -44,13 +45,21 @@ func SearchComplete(ctx context.Context, property string, srchInput *model.Searc
 	if userDataErr != nil {
 		return []*string{}, userDataErr
 	}
+
+	//check that shared cache has resource datatypes:
+	propTypesCache, err := rbac.GetCache().GetPropertyTypes(ctx, false)
```
jlpadilla, line 50: I'd rename `propTypesCache` to `propTypes` or `propTypesMap` because at this point it's no longer part of the cache.

pkg/resolver/searchComplete_test.go
```diff
@@ -98,7 +97,12 @@ func Test_SearchCompleteWithFilter_Query(t *testing.T) {
 	csRes, nsRes, managedClusters := newUserData()
 	ud := rbac.UserData{CsResources: csRes, NsResources: nsRes, ManagedClusters: managedClusters}
 	searchInput := &model.SearchInput{Filters: []*model.SearchFilter{{Property: "namespace", Values: []*string{&value1, &value2}}, {Property: "cluster", Values: []*string{&cluster}}}, Limit: &limit}
-	resolver, mockPool := newMockSearchComplete(t, searchInput, prop1, &ud)
+	PropTypes := make(map[string]string)
+	PropTypes["kind"] = "string"
+	PropTypes["namespace"] = "string"
+	PropTypes["cluster"] = "string"
```
jlpadilla, line 103: Simpler to declare inline
```suggestion
	PropTypes := make(map[string]string) { "kind":"string", "namespace":"string", "cluster":"string" }
```

pkg/resolver/searchComplete_test.go
```diff
@@ -214,11 +218,16 @@ func Test_SearchCompleteWithLabel_Query(t *testing.T) {
 	ud := rbac.UserData{CsResources: csRes, NsResources: nsRes, ManagedClusters: managedClusters}
 	searchInput := &model.SearchInput{Filters: []*model.SearchFilter{{Property: "namespace", Values: []*string{&value1, &value2}}, {Property: "cluster", Values: []*string{&cluster}}}, Limit: &limit}
 
+	// PropTypes := make(map[string]string)
+	// PropTypes["label"] = "object"
+	// PropTypes["namespace"] = "string"
+	// PropTypes["cluster"] = "string"
+
```
jlpadilla, line 225: delete commented code.

pkg/resolver/searchHelper.go
```diff
@@ -0,0 +1,330 @@
...
+	"github.com/doug-martin/goqu/v9"
+	"github.com/doug-martin/goqu/v9/exp"
+	"github.com/lib/pq"
+	"github.com/stolostron/search-v2-api/graph/model"
+	"github.com/stolostron/search-v2-api/pkg/config"
+	"github.com/stolostron/search-v2-api/pkg/rbac"
+	"k8s.io/klog/v2"
+)
+
+func GetPropertyTypeCache(ctx context.Context, refresh bool) (map[string]string, error) {
```
jlpadilla, line 22: Function scope. Most of these functions should be private (lowercase) to the resolver package.  Only make public if the function is needed from a different package.

pkg/resolver/search_test.go
```diff
@@ -40,7 +44,9 @@ func Test_SearchResolver_Count_WithRBAC(t *testing.T) {
 	// Create a SearchResolver instance with a mock connection pool.
 	val1 := "Pod"
 	searchInput := &model.SearchInput{Filters: []*model.SearchFilter{{Property: "kind", Values: []*string{&val1}}}}
-	resolver, mockPool := newMockSearchResolver(t, searchInput, nil, &ud)
+	PropTypes := make(map[string]string)
+	PropTypes["kind"] = "string"
```
jlpadilla, line 48: Can be simplified with inline declaration.  There's several more places where this can be changed below.
```suggestion
	PropTypes := map[string]string { "kind":"string" }
```

pkg/resolver/search_test.go
```diff
@@ -252,10 +264,13 @@ func testAllOperators(t *testing.T, testOperators []TestOperatorItem) {
 	for _, currTest := range testOperators {
 		csRes, nsRes, mc := newUserData()
 		ud := rbac.UserData{CsResources: csRes, NsResources: nsRes, ManagedClusters: mc}
+		PropTypes := make(map[string]string)
+		PropTypes["current"] = "number"
+		PropTypes["created"] = "string"
```
jlpadilla, line 269:
```suggestion
		PropTypes := map[string]string {"current": "number", "created": "string" }
```

jlpadilla: Several more instances below

pkg/resolver/search_test.go
```diff
@@ -12,12 +12,16 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func Test_SearchResolver_Count(t *testing.T) {
+var PropTypes map[string]string
 
+func Test_SearchResolver_Count(t *testing.T) {
+	PropTypes := make(map[string]string)
```
jlpadilla, line 18: - Do we need to declare outside of the test?
- Maybe `propTypeMock` is a better name and would avoid confusion. 
```suggestion

func Test_SearchResolver_Count(t *testing.T) {
	propTypesMock := map[string]string {"kind":"string"}
```

pkg/resolver/testHelper_test.go
```diff
@@ -150,7 +151,8 @@ func newMockRowsWithoutRBAC(mockDataFile string, input *model.SearchInput, prop
 			uid := item.(map[string]interface{})["uid"]
 			cluster := strings.Split(uid.(string), "/")[0]
 			data := item.(map[string]interface{})["properties"].(map[string]interface{})
-
+			// fmt.Println("Property ", data[prop])
+			// fmt.Println("Property Type", reflect.TypeOf(data[prop]))
```
jlpadilla, line 155: Delete

pkg/resolver/testHelper_test.go
```diff
@@ -165,6 +167,7 @@ func newMockRowsWithoutRBAC(mockDataFile string, input *model.SearchInput, prop
 					case map[string]interface{}:
 						propsArray = append(propsArray, v)
 					case []interface{}:
+						// fmt.Println("Reaches interface{} type", v)
```
jlpadilla, line 170: delete

pkg/rbac/sharedData.go
```diff
@@ -51,6 +55,78 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+	if len(cache.shared.propTypes) > 0 && cache.shared.propTypeErr == nil && !refresh {
+		propTypesMap := cache.shared.propTypes
+		return propTypesMap, nil
+
+	} else {
+		klog.V(6).Info("Getting property types from database.")
+		//run query to refresh data
+		propTypes, err := cache.shared.getPropertyTypes(cache, ctx)
+		if err != nil {
+			klog.Errorf("Error retrieving property resources. Error: [%+v]", err)
```
jlpadilla, line 120: Super minor
```suggestion
			klog.Errorf("Error retrieving property types. Error: [%+v]", err)
```

pkg/resolver/search.go
```diff
@@ -157,38 +162,44 @@ func (s *SearchResult) buildSearchQuery(ctx context.Context, count bool, uid boo
...
+					buildRbacWhereClause(ctx, s.userData, userInfo)) // add rbac
+			} else {
+				panic(fmt.Sprintf("RBAC clause is required! None found for search query %+v for user %s ", s.input,
+					ctx.Value(rbac.ContextAuthTokenKey)))
+			}
+		} else if err != nil {
+			klog.Errorf("Error building Search query: %s", err.Error())
 		}
+	} else {
+		klog.Errorf("Error building Search query: %s", err.Error())
```
jlpadilla, line 195: Please update this log.
```suggestion
		klog.Errorf("Error: query input must contain a filter or keyword. Received: %+v", s.input)
```

jlpadilla: Resolved with this commit. https://github.com/stolostron/search-v2-api/pull/91/commits/1666e7a9398314d715ba987f746e1f245ceffcf9

pkg/rbac/sharedData.go
jlpadilla: Resolved with this commit. https://github.com/stolostron/search-v2-api/pull/91/commits/1666e7a9398314d715ba987f746e1f245ceffcf9

jlpadilla (approved):
/lgtm
Thank you @Anxhela21 !

pkg/rbac/sharedData.go
```diff
@@ -51,39 +56,119 @@ var managedClusterResourceGvr = schema.GroupVersionResource{
...
+			continue
+		}
+		propTypeMap[key] = value
+
+	}
+	//We need to manually add the cluster proptype as string because it does not appear in the data column:
+	// NOTE: we will have to do this for any property we choose to remove from data field and to column.
+	propTypeMap["cluster"] = "string"
+
+	fmt.Println(propTypeMap)
```
SherinV, line 106: remove

pkg/resolver/related_test.go
```diff
@@ -104,7 +106,7 @@ func Test_SearchResolver_RelatedKindsRelationships(t *testing.T) {
 	config.Cfg.RelationLevel = 3
 
 	var resultList []*string
-
+	propTypesMock := map[string]string{"uid": "string"}
```
SherinV, line 109: Does the test need uid to be in the map?

pkg/resolver/search.go
```diff
@@ -509,41 +293,46 @@ func WhereClauseFilter(input *model.SearchInput) []exp.Expression {
...
+							klog.Warningf("Error creating property type map with err: [%s] ", err)
 						}
+						propTypeMap = propTypeMapNew
+						break
 
-					}
-					values = cleanedVal
-				}
+					} else {
+						klog.V(5).Infof("Prop in map:%s, filter prop is: %s, datatype :%s\n", dataTypeInMap, filter.Property)
```
SherinV, line 314:
```suggestion
						klog.V(5).Infof("For filter prop: %s, datatype is :%s\n", filter.Property, dataTypeInMap)
```
Remove the extra %s.

SherinV:
/lgtm
Added a few minor comments.

pkg/resolver/related_test.go
Anxhela21: I don't think it's necessary for these tests. The related resources does not directly depend on property types (only initial query does)

Anxhela21: Removing from tests.

pkg/resolver/search.go
Anxhela21: that's more clear! ty!

PR #93, opened 2022-09-21
Add tests for messages

SherinV:

### Description of changes
- Add tests for messages.go


pkg/rbac/messagesDev.go
```diff
@@ -0,0 +1,65 @@
+// Copyright Contributors to the Open Cluster Management project
```
jlpadilla, line 1: Can we use `_test.go` in the file name?

SherinV: I am getting `pkg/resolver/messages_test.go:20:25: undefined: rbac.NewMockCacheForMessages` error if I do that.

jlpadilla:
This PR was replaced by #95

PR #98, opened 2022-09-29
Avoid leaking properties in loop

SherinV:

**Related Issue:**  https://github.com/stolostron/backlog/issues/26006

### Description of changes
To avoid the issue we saw in PM cluster where Pods had apigroup set. Since data was defined outside, it leaked properties while looping. Moving the declaration inside. @zlayne

### DIFF
--- a/pkg/resolver/search.go
+++ b/pkg/resolver/search.go
@@ -259,12 +259,12 @@ func (s *SearchResult) resolveItems() ([]map[string]interface{}, error) {
 	}
 	defer rows.Close()
 
-	var cluster string
-	var data map[string]interface{}
 	s.uids = make([]*string, len(items))
 
 	for rows.Next() {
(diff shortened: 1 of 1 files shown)


jlpadilla (approved):
/lgtm

🥇 Thank you! I incorrectly assumed that values would get "reset" by Scan().

jlpadilla:
@SherinV I wonder if this will also resolve this problem with `_hostingSubscription` being added to resources that shouldn't have it? https://github.com/stolostron/backlog/issues/26006

SherinV:
> @SherinV I wonder if this will also resolve this problem with `_hostingSubscription` being added to resources that shouldn't have it? [stolostron/backlog#26006](https://github.com/stolostron/backlog/issues/26006)

Hmm, that's true. Might be coming from here.

PR #99, opened 2022-09-30
remove kubeadmin workaround

SherinV:

**Related Issue:**  [stolostron/backlog#25996](https://github.com/stolostron/backlog/issues/25996)

### Description of changes
Remove rbac workaround for kubeadmin
For users to get the correct SSRR, all provided user info should be passed to impersonation config.
Just passing username and UID will not suffice to create the correct SSRR, if other info like groups and extra are present in token review.

### DIFF
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -320,30 +320,30 @@ func (shared *SharedData) isClusterScoped(kindPlural, apigroup string) bool {
 	return ok
 }
 
-func setImpersonationUserInfo(user authv1.UserInfo) *rest.ImpersonationConfig {
+func setImpersonationUserInfo(userInfo authv1.UserInfo) *rest.ImpersonationConfig {
 	impersonConfig := &rest.ImpersonationConfig{}
 	// All fields in user info, if set, should be added to ImpersonationConfig. Otherwise SSRR won't work.
 	// All fields in UserInfo is optional. Set only if there is a value
 	//set username
-	if user.Username != "" {
-		impersonConfig.UserName = user.Username
+	if userInfo.Username != "" {
+		impersonConfig.UserName = userInfo.Username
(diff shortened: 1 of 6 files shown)


jlpadilla (approved):
/lgtm
/hold

pkg/rbac/userData.go
```diff
@@ -258,6 +258,9 @@ func (user *UserDataCache) getNamespacedResources(cache *Cache, ctx context.Cont
 		} else {
 			klog.V(9).Infof("SelfSubjectRulesReviews Kube API result for ns:%s : %v\n", ns, prettyPrint(result.Status))
 		}
+
+		// Name the loop - to break out of the loop if user is kubeadmin (have access to everything)
```
jlpadilla, line 262: The user doesn't have to be 'kubeadmin', other users could have similar `*` access.
```suggestion
		// Name the loop - to break out of the loop if user has access to everything (*)
```

SherinV: Updated

pkg/rbac/userData_test.go
```diff
@@ -610,3 +611,81 @@ func Test_getUserData(t *testing.T) {
...
+		userData:     UserData{},
+		nsrUpdatedAt: time.Now(),
+	}
+	_, err := udc.getImpersonationClientSet("123456", mock_cache)
+	// Ensure that there is no error
+	assert.Nil(t, err)
+
+}
+
+func Test_getNamespaces_kubeAdmin(t *testing.T) {
```
jlpadilla, line 646: Rename test.
```suggestion
// User has access to all resources in the namespace.
func Test_hasAccessToAllResourcesInNamespace(t *testing.T) {
```

jlpadilla (approved):
/lgtm

PR #100, opened 2022-10-03
Case insensitive keywords

Anxhela21:

### Description of changes
Users should be able to search by keyword without worrying about case sensitivity. This PR fixes current logic to use a case insensitive like clause


jlpadilla (approved):
/lgtm

This matches the behavior in v1.  We'll have to watch for the performance impact.

PR #101, opened 2022-10-04
update user's managedcluster list

SherinV:

**Related Issue:**  stolostron/backlog#25996

### Description of changes
Update user's managed clusters too if user has access to that namespace
Correct logs without formatting in the Infof.


jlpadilla (approved):
/lgtm
/hold

PR #102, opened 2022-10-05
check if user has access to all resources

SherinV:

**Related Issue:**  stolostron/backlog#22013 - Optimize RBAC for kubeadmin level access

### Description of changes
Before creating SSRR and SSAR for each namespace and resource - check if user has kubeadmin level access/access to all resources. This will bypass ns level checks and make it easier for kubeadmin level  users.

### DIFF
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -232,10 +232,9 @@ func (user *UserDataCache) userAuthorizedListSSAR(ctx context.Context, authzClie
 	accessCheck := &authz.SelfSubjectAccessReview{
 		Spec: authz.SelfSubjectAccessReviewSpec{
 			ResourceAttributes: &authz.ResourceAttributes{
-				Namespace: "*",
-				Verb:      "list",
-				Group:     apigroup,
-				Resource:  kind_plural,
+				Verb:     "list",
+				Group:    apigroup,
+				Resource: kind_plural,
 			},
 		},
 	}
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -74,9 +74,9 @@ func (cache *Cache) GetUserDataCache(ctx context.Context,
 	var user *UserDataCache
 	var uid string
 	var err error
-
+	var userInfo authv1.UserInfo
 	// get uid from tokenreview
-	if uid, _ = cache.GetUserUID(ctx); uid == "noUidFound" {
+	if uid, userInfo = cache.GetUserUID(ctx); uid == "noUidFound" {
 		return user, fmt.Errorf("cannot find user with uid: %s", uid)
 	}
 	clientToken := ctx.Value(ContextAuthTokenKey).(string)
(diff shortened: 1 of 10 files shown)


pkg/rbac/userData.go
```diff
@@ -108,6 +110,19 @@ func (cache *Cache) GetUserDataCache(ctx context.Context,
 		}
 	}
 
+	// Before checking each namespace and clusterscoped resource, check if user has access to everything
+	userHasAllAccess, err := user.userHasAllAccess(cache, ctx, clientToken)
+	if err != nil {
+		klog.Warning("Encountered error while checking if user has access to everything ", err)
+	} else {
+		if userHasAllAccess {
+			klog.Infof("User %s with uid %s has access to all resources.", userInfo.Username, userInfo.UID)
```
xrajesh, line 119: Do we want to log this every time ? V(3) may be ?

pkg/rbac/userData.go
```diff
@@ -108,6 +110,19 @@ func (cache *Cache) GetUserDataCache(ctx context.Context,
...
+			return user, nil
+		}
+		klog.Infof("User %s with uid %s doesn't have access to all resources. Checking individually",
+			userInfo.Username, userInfo.UID)
```
xrajesh, line 123: Same like above , do we need to log every time ?

SherinV: Updated.

pkg/rbac/userData.go
```diff
@@ -186,14 +227,15 @@ func (user *UserDataCache) getClusterScopedResources(cache *Cache, ctx context.C
...
-func (user *UserDataCache) userAuthorizedListCSResource(ctx context.Context, authzClient v1.AuthorizationV1Interface,
+func (user *UserDataCache) userAuthorizedListSSAR(ctx context.Context, authzClient v1.AuthorizationV1Interface,
 	apigroup string, kind_plural string) bool {
 	accessCheck := &authz.SelfSubjectAccessReview{
 		Spec: authz.SelfSubjectAccessReviewSpec{
 			ResourceAttributes: &authz.ResourceAttributes{
-				Verb:     "list",
-				Group:    apigroup,
-				Resource: kind_plural,
+				Namespace: "*",
```
xrajesh, line 235: Do we need this `Namespace: "*" ` Doc says empty "" for all namespace scoped resources .
```
Namespace is the namespace of the action being requested. Currently, there is no distinction between no namespace and all namespaces "" (empty) is defaulted for LocalSubjectAccessReviews "" (empty) is empty for cluster-scoped resources "" (empty) means "all" for namespace scoped resources from a SubjectAccessReview or SelfSubjectAccessReview```
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#resourceattributes-v1-authorization-k8s-io

SherinV: Ok, will remove that. Thanks.

SherinV: Done.

xrajesh (approved):

PR #110, opened 2022-10-24
Workaround: Test the database connection with each request.

jlpadilla:
**Related Issue:**  https://github.com/stolostron/backlog/issues/26447

### Description of changes
Want to merge this workaround to unblock the canary while we find a better solution.


SherinV (approved):

pkg/rbac/cache.go
```diff
@@ -43,5 +43,10 @@ var cacheInst = Cache{
 
 // Get a reference to the cache instance.
 func GetCache() *Cache {
+	// Workaround. Update cache connection with every request.
+	// We need a better way to maintain this connection.
+	cacheInst.pool = db.GetConnection()
+	cacheInst.shared.pool = db.GetConnection()
```
Anxhela21, line 49: why do we need to reference two pools in cache?

jlpadilla: This is because of the change I made in PR #104 
I made the `SharedData` an independent struct with the goal that tests don't need to mock the entire Cache object.
However, the side effect was this duplication of the `pool` object and of course the canary problem.
Let's discuss tomorrow.

PR #111, opened 2022-10-27
Parallelize shared cache and add func to log slow queries

jlpadilla:
**Related Issue:**  https://github.com/stolostron/backlog/issues/22013

### Description of changes
- Parallelize shared cache API requests.
- Add func to log slow queries

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -30,6 +30,7 @@ type Config struct {
 	PlaygroundMode bool // Enable the GraphQL Playground client.
 	QueryLimit     int  // The default LIMIT to use on queries. Client can override.
 	RelationLevel  int  // The number of levels/hops for finding relationships for a particular resource
+	SlowLog        int  // Logs performance data for queries slower than the specified time duration in ms.
 	// Placeholder for future use.
 	// QueryLoopLimit          int // number of queries handled at a time
 	// RBAC_INACTIVITY_TIMEOUT int
@@ -52,6 +53,7 @@ func new() *Config {
 		HttpPort:       getEnvAsInt("HTTP_PORT", 4010),
 		PlaygroundMode: getEnvAsBool("PLAYGROUND_MODE", false),
 		QueryLimit:     getEnvAsInt("QUERY_LIMIT", 1000),
+		SlowLog:        getEnvAsInt("SLOW_LOG", int(500)),
 		//Setting default level to 0 to check if user has explicitly set this variable
 		// This will be updated to 1 for default searches and 3 for applications - unless set by the user
 		RelationLevel: getEnvAsInt("RELATION_LEVEL", 0),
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
```diff
@@ -52,6 +53,7 @@ func new() *Config {
 		HttpPort:       getEnvAsInt("HTTP_PORT", 4010),
 		PlaygroundMode: getEnvAsBool("PLAYGROUND_MODE", false),
 		QueryLimit:     getEnvAsInt("QUERY_LIMIT", 1000),
+		SlowLog:        getEnvAsInt("SLOW_LOG", int(300)),
```
Anxhela21, line 56: Why are we setting default to 300 ms? Will we adjust this once we have better idea of funcs average run time?

pkg/rbac/sharedData.go
```diff
@@ -140,39 +141,46 @@ func (cache *Cache) GetPropertyTypes(ctx context.Context, refresh bool) (map[str
 	}
 }
 
-func (cache *Cache) PopulateSharedCache(ctx context.Context) error {
+func (cache *Cache) PopulateSharedCache(ctx context.Context) {
+	defer metric.SlowLog("PopulateSharedCache", 200*time.Millisecond)()
```
Anxhela21, line 145: should this be 300*time.Millisecond?

pkg/config/config.go
jlpadilla: Yes, 300ms is my initial guess, we'll adjust as we learn more. 
I also added an env variable `SLOW_LOG` in case we need to adjust for a specific cluster.

pkg/rbac/sharedData.go
jlpadilla: In this case it is overriding the default (300ms) for this specific function to be lower (200ms).
With this we can set a more accurate SlowLog based on what each function is doing.

pkg/rbac/sharedData.go
```diff
@@ -140,39 +141,46 @@ func (cache *Cache) GetPropertyTypes(ctx context.Context, refresh bool) (map[str
...
 		klog.V(5).Info("Using shared data from cache.")
-		return nil
-	} else { // get data and cache
+		return
+	} else { // get data and add to cache
+		var wg sync.WaitGroup
+
+		// get hub cluster-scoped resources and cache in shared.csResources
+		wg.Add(1)
+		go func() {
```
xrajesh, line 154: These go-routines could have a timeout ? we may be spinning off go-routines, and they will be making many Kube-api calls , even after the client request is dead.

jlpadilla: Yes, the correct way to handle timeouts is with the `context` object.
The kube and postgres clients both accept a context. We just need to pass it down from the request. We have been steadily improving oour use of context, but we aren't perfect yet.

Anxhela21 (approved):
/lgtm

PR #113, opened 2022-11-01
Parallelize SSRR requests for namespaces

jlpadilla:

### Description of changes
- Use an async go routine for each namespace SSRR request.

### DIFF
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -215,7 +215,7 @@ func (user *UserDataCache) getClusterScopedResources(cache *Cache, ctx context.C
 	// If we have a new set of authorized list for the user reset the previous one
 	user.userData.CsResources = nil
 
-	// Paralellize API calls.
+	// Paralellize SSAR API calls.
 	var wg sync.WaitGroup
 	var lock sync.Mutex
 	// For each cluster-scoped resource, check if the user is authorized to list.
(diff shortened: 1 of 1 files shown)


pkg/rbac/userData.go
```diff
@@ -277,10 +277,82 @@ func (user *UserDataCache) updateUserManagedClusterList(cache *Cache, ns string)
...
+		klog.V(9).Infof("SelfSubjectRulesReviews Kube API result for ns:%s : %v\n", ns, prettyPrint(result.Status))
+	}
+
+	// Process the SSRR result and add to this UserDataCache object.
+	for _, rules := range result.Status.ResourceRules {
+		for _, verb := range rules.Verbs {
+			if verb == "list" || verb == "*" {
+				for _, res := range rules.Resources {
+					for _, api := range rules.APIGroups {
+						lock.Lock()
```
Anxhela21, line 305: Wondering would it be better to lock after we find resource with verb == "list" || verb ="*" (third line in loop)?

jlpadilla: Maybe. I don't have a definite answer but here are my thoughts on the pros and cons.
- Here we lock for the shortest period of time. This creates a large number of lock/unlock cycles which allows other go routines to get the lock and continue their processing.
- Moving to the outside loop we have less lock/unlock cycles, which requires less computation, but the hold the lock for a longer period blocking other go routines.

Anxhela21: these are good points. we can revisit later as potential optimization.

PR #117, opened 2022-11-02
Refactor namespaces cache and expand use of generic cacheMetadata

jlpadilla:
**Related Issue:**  N/A

### Description of changes
- Use common `isValid()` function for cached data.
- Use function to access namespaces from shared cache.
- Cache errors for a shorter time duration. (TTL = 1 second)


pkg/rbac/cacheMetadata.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/stolostron/search-v2-api/pkg/config"
+)
+
+// Common fields to manage a cached data field.
+type cacheMetadata struct {
+	err       error         // Error while retrieving the data from external API.
+	lock      sync.Mutex    // Locks the data field while requesting the latest data.
+	updatedAt time.Time     // Time when the data field was last updated.
+	ttl       time.Duration // Time-to-live, time duration for which this cache is valid.
```
SherinV, line 16: will ttl be different for different caches ( namespace vs clusterscoped vs ..)going forward?

SherinV: Assuming this field will be set in the upcoming PRs.

SherinV (approved):

pkg/rbac/cacheMetadata.go
Anxhela21: is the idea that we are moving it out of config.go or will this point to it?

Anxhela21: currently, default ttl values differ for token review and shared/user data cache. But we still need to adjust those times to something more realistic.

Anxhela21 (approved):
/lgtm

pkg/rbac/cacheMetadata.go
jlpadilla: Right, ttl is intended to be adjusted for different type of data (token review, shared, user).  This PR is limiting the scope to sharedData, but it will get more use in my next change.

PR #121, opened 2022-12-13
Update the RBAC cache when Namespaces or ManagedClusters change

jlpadilla:
**Related Issue:**  https://issues.redhat.com/browse/ACM-1811

### Description of changes
- Watch Namespaces and update the RBAC cache.
- Watch ManagedClusters and update RBAC cache.

### To test this change
Setup
```
export ns=test1
export usr=sa2
oc create serviceaccount ${usr} -n ${ns}
oc create role ${usr} --verb=list --resource=configmaps -n ${ns}
oc create rolebinding ${usr} --role=${usr} --serviceaccount=${ns}:${usr} -n ${ns}
oc create clusterrole ${usr} --verb=list --resource=nodes,configmaps -n ${ns}
oc create clusterrolebinding ${usr} --clusterrole=${usr} --serviceaccount=${ns}:${usr} -n ${ns}
export TOKEN=$(oc create token ${usr} -n ${ns})
export URL=https://localhost:4010/searchapi/graphql  
```

Create a new namespace and search for resources on that namespace shortly after creating it.
```
oc create namespace test-ns
sleep 6
curl --insecure --location \
--request POST $URL \
--header "Authorization: Bearer $TOKEN" --header 'Content-Type: application/json' \
--data-raw '{"query":"query q($input: [SearchInput]) { search(input: $input) { items } }","variables":{"input":[{"keywords":[],"filters":[{"property":"namespace","values":["test-ns"]}], "limit":10}]}}' | jq
```


pkg/rbac/cacheValidation.go
```diff
@@ -0,0 +1,213 @@
...
+	c.shared.nsCache.lock.Unlock()
+
+	// Add the namespace to each user in UserData cache.
+	c.usersLock.Lock()
+	defer c.usersLock.Unlock()
+	wg := sync.WaitGroup{}
+	lock := sync.Mutex{}
+	for _, userCache := range c.users {
+		wg.Add(1)
+		go func(userCache *UserDataCache) { // All users updated asynchhronously
```
SherinV, line 136:
```suggestion
		go func(userCache *UserDataCache) { // All users updated asynchronously
```

pkg/rbac/cacheValidation.go
```diff
@@ -0,0 +1,213 @@
...
+				return
+
+			case event := <-watch.ResultChan(): // Read events from the watch channel.
+				klog.V(6).Infof("Event: %s \tResource: %s  ", event.Type, w.gvr.String())
+				o, err := runtime.UnstructuredConverter.ToUnstructured(runtime.DefaultUnstructuredConverter, &event.Object)
+				if err != nil {
+					klog.Warningf("Error converting %s event.Object to unstructured.Unstructured. Error: %s",
+						w.gvr.Resource, err)
+				}
+				obj := &unstructured.Unstructured{Object: o}
```
SherinV, line 95: Can we convert to unstructured if error happens in the above line (if err!=nil)?

pkg/rbac/cacheValidation.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Update the cache when a namespace is DELETED.
+func (c *Cache) namespaceDeleted(obj *unstructured.Unstructured) {
+	// Delete from Namespaces shared cache.
+	c.shared.nsCache.lock.Lock()
+	defer c.shared.nsCache.lock.Unlock()
+	ns := obj.GetName()
+	newNamespaces := make([]string, 0)
+	for _, n := range c.shared.namespaces {
```
SherinV, line 154: Is it possible to convert namespaces also to a map structure so that we don't have loop through it to delete?

SherinV (approved):
Minor comments only. /lgtm

pkg/rbac/cacheValidation.go
jlpadilla: Yes! I want to change it also for consistency with the other data, however that's a big change, so I'll keep it separate from this PR.

pkg/rbac/cacheValidation.go
jlpadilla: If there's an error above, I assume that `o` will be an empty `map[string]struct{}` which could still be converted to unstructured, it will just be an empty object.

jlpadilla: So the error could potentially cause problems in the logic after, but II'm worried that returning could cause more harm.

pkg/rbac/cacheValidation.go
```diff
@@ -0,0 +1,213 @@
...
+					time.Sleep(5 * time.Second)
+				}
+			}
+		}
+	}
+}
+
+// Update the cache when a namespace is ADDED.
+func (c *Cache) namespaceAdded(obj *unstructured.Unstructured) {
+	// Addd namespace to shared cache.
```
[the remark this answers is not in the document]
jlpadilla: done

Anxhela21 (approved):
/lgtm
/hold

PR #143, opened 2023-03-14
ACM-3288 Exclude scraping requests from prom data

jlpadilla:
### Related Issue

https://issues.redhat.com/browse/ACM-3288

### Description of changes
- Use a custom prometheus registry to exclude scraping requests (/metrics) from the returned data.
- Rename metric
- Add Make target to setup database for local dev.

### DIFF
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -7,22 +7,33 @@ import (
 )
 
 var (
-	HttpDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
-		Name: "search_http_duration_seconds",
-		Help: "Latency of of HTTP requests in seconds.",
+	HttpRequestsHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "search_api_requests",
+		Help: "Histogram of HTTP requests duration (seconds).",
 	}, []string{"code"})
 )
 
 var (
 	DBConnectionFailed = promauto.NewCounterVec(prometheus.CounterOpts{
-		Name: "search_db_connection_failed_total",
+		Name: "search_api_db_connection_failed_total",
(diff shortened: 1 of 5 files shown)


pkg/metric/metric.go
```diff
@@ -7,22 +7,33 @@ import (
...
 	DBQueryDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
-		Name: "search_dbquery_duration_seconds",
+		Name: "search_api_dbquery_duration_seconds",
 		Help: "Latency of DB requests in seconds.",
 	}, []string{"query_name"})
 )
+
+// var (
+// 	promRegistry = prometheus.NewRegistry()
+// )
```
Anxhela21, line 32: remove commented code

pkg/metric/metric.go
```diff
@@ -7,22 +7,33 @@ import (
...
+
+func GetPromRegistry() *prometheus.Registry {
+	r := prometheus.NewRegistry()
+	r.MustRegister(HttpRequestsHistogram)
+
+	return r
+}
```
Anxhela21, line 39: promauto will automatically register the metric when we create the histogram using that package. wondering if doing it again causes any issues?

pkg/metric/metric.go
jlpadilla: done

pkg/metric/metric.go
jlpadilla: There's a default registry, which is what's creating the problems. The default registry has other metrics like go_* metrics which seem to add unnecessary overhead.

With the change here I create a custom registry, it only has our 3 metrics. That seems to work better but I'm not aware if it can cause other issues.

I pushed a commit to simplify this change.

Anxhela21 (approved):
/lgtm
/hold

PR #148, opened 2023-04-05
lock map

SherinV:
### Related Issue

### Description of changes
- Added ...

### DIFF
--- a/pkg/resolver/rbacHelper.go
+++ b/pkg/resolver/rbacHelper.go
@@ -3,6 +3,7 @@ package resolver
 
 import (
 	"encoding/json"
+	"sync"
 
 	"github.com/doug-martin/goqu/v9"
 	"github.com/doug-martin/goqu/v9/exp"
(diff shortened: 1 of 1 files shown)


SherinV:
Looks like the error happens on user.NsResources.

pkg/resolver/rbacHelper.go
```diff
@@ -134,15 +135,23 @@ func matchNamespacedResources(nsResources map[string][]rbac.Resource, userInfo v
...
+			mMutex.RUnlock()
+
+			// Lock map before write
+			mMutex.Lock()
+			if found {
 				m[string(b)] = append(m[string(b)], ns)
 			} else {
 				m[string(b)] = []string{ns}
 			}
+			mMutex.Unlock()
```
jlpadilla, line 154: The lock might cause a problem when reading inside the `if{}`.  This should be safer.
```suggestion
			mMutex.RLock()
			prevValue, found := m[string(b)]
			mMutex.RUnlock()

			// Lock map before write
			mMutex.Lock()
			if found {
				m[string(b)] = append(prevValue, ns)
			} else {
				m[string(b)] = []string{ns}
			}
			mMutex.Unlock()
```

SherinV: I have removed this lock for now. Don't think this is the one causing the error.

pkg/resolver/rbacHelper.go
```diff
@@ -138,7 +138,10 @@ func consolidateNsResources(nsResources map[string][]rbac.Resource) (map[string]
 	for ns, resources := range nsResources {
 		b, err := json.Marshal(resources)
 		if err == nil {
-			if _, found := m[string(b)]; found {
+			_, found := m[string(b)]
+
+			// Lock map before write
+			if found {
```
jlpadilla, line 144: Revert this change.

SherinV: Done

pkg/rbac/userData.go
```diff
@@ -307,7 +307,9 @@ func (user *UserDataCache) getSSRRforNamespace(ctx context.Context, cache *Cache
 							// No need to loop through all resources. Save the wildcard *
 							// exit the resourceRulesLoop
 							if res == "*" && api == "*" {
+								// user.nsrCache.lock.Lock()
```
jlpadilla, line 310: Why are we removing this lock?
I think it's still needed because it's possible to have multiple requests from the same user. Maybe we could simplify by locking at a higher level.

SherinV: Adding that causes the rbac.(*UserDataCache).getNamespacedResources function to timeout and fail in Test_getNamespaces_emptyCache

jlpadilla (approved):
/lgtm
/hold

PR #173, opened 2023-08-10
ACM-6165: Include property with list of relatedUids in related objects

SherinV:
### Related Issue
https://issues.redhat.com/browse/ACM-6165

### Description of changes
- Related objects include a new property `_relatedUids` with a list of objects it is related to.


jlpadilla (changes requested):
@philipwu08 @SherinV I want to replace the words "parent" and "child" in this PR.  Not all relations are parent-child, it is only one type of relationship.
To avoid confusion, let's use something more generic like `relatedUid`.

SherinV:
> 

Replaced

pkg/resolver/related.go
```diff
@@ -342,6 +357,39 @@ func (s *SearchResult) updateKindMap(uid string, kind string, levelMap map[strin
 	levelMap[kind] = uids
 }
 
+// map the related result uids to the uids in the search input
```
jlpadilla, line 360:
```suggestion
// Maps the related result uids to the uids in the search result set.
```

pkg/resolver/related.go
```diff
@@ -342,6 +357,39 @@ func (s *SearchResult) updateKindMap(uid string, kind string, levelMap map[strin
...
+				}
+			}
+			klog.V(9).Infof("uid %s is  related to uids %+v.", resultUid, resultToCurrSearchUidsMap[resultUid])
+			// Need to process only one uid in the path as it is either the sourceid or the destid that can be related
+			break
+		}
+	}
+}
+
+func checkIfInArray(lookupMap []string, uid string) bool {
```
jlpadilla, line 384: Just FYI, go 1.21 will add a new standard package `slices` to simplify this. It also has a `slices.Sort()` which we have somewhere else in this project.
https://pkg.go.dev/slices

jlpadilla (approved):
/lgtm
/hold

PR #174, opened 2023-08-29
ACM-7062: Special RBAC handling for global search user

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-7062

### Description of changes
- Implement the role `searches/allManagedData` which grants the global search user access to all resources from managed clusters.

### DIFF
--- a/pkg/rbac/userData.go
+++ b/pkg/rbac/userData.go
@@ -156,7 +156,7 @@ func (user *UserDataCache) userHasAllAccess(ctx context.Context, cache *Cache) (
 		user.ManagedClusters = map[string]struct{}{"*": {}}
 		user.clustersCache.updatedAt = time.Now()
 		user.csrCache.err, user.nsrCache.err, user.clustersCache.err = nil, nil, nil
-		klog.V(5).Infof("User %s with uid %s has *list* access to all managed cluster resources.",
+		klog.V(5).Infof("User %s with uid %s has access to all resources.",
 			user.userInfo.Username, user.userInfo.UID)
 
 		return true, nil
@@ -177,7 +177,7 @@ func (user *UserDataCache) userHasAllAccess(ctx context.Context, cache *Cache) (
 		user.ManagedClusters = map[string]struct{}{"*": {}}
 		user.clustersCache.updatedAt = time.Now()
 		user.csrCache.err, user.nsrCache.err, user.clustersCache.err = nil, nil, nil
-		klog.V(5).Infof("User %s with uid %s has *get* access to all managed cluster resources.",
+		klog.V(5).Infof("User %s with uid %s is authorized to search/allManagedData which gives access to all managed cluster resources.",
 			user.userInfo.Username, user.userInfo.UID)
 		return true, nil
 	}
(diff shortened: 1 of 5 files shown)


pkg/rbac/userData.go
```diff
@@ -153,9 +153,32 @@ func (user *UserDataCache) userHasAllAccess(ctx context.Context, cache *Cache) (
...
+		defer user.nsrCache.lock.Unlock()
+		user.NsResources = map[string][]Resource{}
+		user.nsrCache.updatedAt = time.Now()
+
+		cache.shared.mcCache.lock.Lock()
+		defer cache.shared.mcCache.lock.Unlock()
+		user.ManagedClusters = map[string]struct{}{"*": {}}
+		user.clustersCache.updatedAt = time.Now()
+		user.csrCache.err, user.nsrCache.err, user.clustersCache.err = nil, nil, nil
+		klog.V(5).Infof("User %s with uid %s has *get* access to all managed cluster resources.",
```
jlpadilla, line 180: What's the difference between list and get? 
Maybe we should log "user is authorized to search/allManagedData which gives access to a all managed cluster resources."

SherinV: Used 'get' as we gave that permission in the cluster role here. https://github.com/stolostron/search-v2-operator/blob/7f9bc36bd8a5dc35863c35b73dc43b91737d74db/controllers/create_rolesbindings.go#L214
'get' gets a resource by name while list lists all based on the selector, is what I think.  @jlpadilla

pkg/rbac/userData.go
```diff
@@ -153,9 +153,32 @@ func (user *UserDataCache) userHasAllAccess(ctx context.Context, cache *Cache) (
...
-		user.ManagedClusters = cache.shared.managedClusters
+		user.ManagedClusters = map[string]struct{}{"*": {}}
 		user.clustersCache.updatedAt = time.Now()
 		user.csrCache.err, user.nsrCache.err, user.clustersCache.err = nil, nil, nil
+		klog.V(5).Infof("User %s with uid %s has access to all resources.",
+			user.userInfo.Username, user.userInfo.UID)
+
+		return true, nil
+	} else if user.userAuthorizedListSSAR(ctx, impersClientSet,
+		"get", "searches", "searches/allManagedData") {
```
jlpadilla, line 164: **Future optimization**
This if/else requires 2 client requests which are executed synchronously. We could save a few milliseconds if we execute these asynchronously.

pkg/rbac/userData.go
```diff
@@ -153,9 +153,32 @@ func (user *UserDataCache) userHasAllAccess(ctx context.Context, cache *Cache) (
...
+		user.csrCache.lock.Lock()
+		defer user.csrCache.lock.Unlock()
+		user.CsResources = []Resource{}
```
jlpadilla, line 167: **Only for discussion, no changes needed from this comment**

When the user is authorized to `searches/allManagedData` it means that they won't get any local resources even if authorized. 
This won't help the ACM Admin scenario as I hoped.  If we want to cover this scenario in the future, we could do it with a new subresource like `searches/inherit-local-bypass-managed`, which would implement this new behavior.

**The current implementation optimizes the Global Search user scenario which is all we need right now.**

jlpadilla (approved):
/lgtm
/hold

PR #175, opened 2023-09-01
use jsonb operator in where clause

SherinV:
### Related Issue

### Description of changes
- Added ...

### DIFF
--- a/pkg/rbac/sharedData.go
+++ b/pkg/rbac/sharedData.go
@@ -199,7 +199,7 @@ func (shared *SharedData) getClusterScopedResources(ctx context.Context) error {
 
 	// Building query to get cluster scoped resources
 	// Original query: "SELECT DISTINCT data->>'apigroup', data->>'kind_plural' FROM search.resources WHERE
-	// data->>'_hubClusterResource'='true' AND data->>'namespace' is NULL"
+	// data?'_hubClusterResource' AND data?'namespace' is FALSE"
 	schemaTable := goqu.S("search").Table("resources")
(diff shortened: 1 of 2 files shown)


pkg/rbac/sharedData.go
```diff
@@ -202,7 +204,8 @@ func (shared *SharedData) getClusterScopedResources(ctx context.Context) error {
 	ds := goqu.From(schemaTable)
 	query, _, err := ds.SelectDistinct(goqu.COALESCE(goqu.L(`"data"->>'apigroup'`), "").As("apigroup"),
 		goqu.COALESCE(goqu.L(`"data"->>'kind_plural'`), "").As("kind")).
-		Where(goqu.L(`"data"->>'_hubClusterResource'='true'`), goqu.L(`"data"->>'namespace'`).IsNull()).ToSQL()
+		Where(goqu.L("???", goqu.C("data"), goqu.Literal("?"), "_hubClusterResource"),
+			goqu.L("???", goqu.C("data"), goqu.Literal("?"), "namespace").IsFalse()).ToSQL()
```
jlpadilla, line 208: Just to confirm, I assume `IsFalse()` covers the case of NULL?
Does this change the sample query in the comment above?

jlpadilla: Is there a performance benefit? 
For the namespace part `IsNull()` seemed more intuitive.I'm ok changing if it helps in any other way.

jlpadilla (approved):
/lgtm
/hold

PR #177, opened 2023-09-13
fix crash loop

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-7467(https://issues.redhat.com/browse/ACM-7467)

### Description of changes
- Added ...

### DIFF
--- a/pkg/rbac/authzMiddleware.go
+++ b/pkg/rbac/authzMiddleware.go
@@ -8,11 +8,6 @@ import (
 
 func AuthorizeUser(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// if postgres db is not setup, return error
-		if !GetCache().dbConnInitialized {
-			http.Error(w, "Unable to establish connection with Postgres.", 500)
-			return
-		}
 
 		// Trigger initialization of the shared cache. We should move this to a
 		// different place where it's independent of the request.
--- a/pkg/rbac/authzMiddleware.go
+++ b/pkg/rbac/authzMiddleware.go
@@ -8,6 +8,11 @@ import (
 
 func AuthorizeUser(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// if postgres db is not setup, return error
+		if !GetCache().dbConnInitialized {
+			http.Error(w, "Unable to establish connection with Postgres.", 500)
+			return
+		}
 
 		// Trigger initialization of the shared cache. We should move this to a
 		// different place where it's independent of the request.
--- a/pkg/rbac/cache.go
+++ b/pkg/rbac/cache.go
@@ -33,7 +33,7 @@ func (c *Cache) GetDbConnInitialized() bool {
 	return c.dbConnInitialized
 }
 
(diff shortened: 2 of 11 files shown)


pkg/rbac/authzMiddleware.go
```diff
@@ -8,6 +8,11 @@ import (
 
 func AuthorizeUser(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// if postgres db is not setup, return error
+		if !GetCache().dbConnInitialized {
+			http.Error(w, "Unable to establish connection with Postgres.", 500)
```
jlpadilla, line 13: Maybe `503 Service Unavailable` is a little more specific.

pkg/rbac/authzMiddleware.go
```diff
@@ -8,6 +8,11 @@ import (
 
 func AuthorizeUser(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// if postgres db is not setup, return error
```
jlpadilla, line 11: Maybe this should be a separate middleware.  It's a small check, but not required for authorization.

pkg/rbac/cache.go
```diff
@@ -22,9 +23,18 @@ type Cache struct {
...
+	pool              pgxpoolmock.PgxPool // Database client
+	restConfig        *rest.Config
+	dbConnInitialized bool
+}
+
+func (c *Cache) GetDbConnInitialized() bool {
+	return c.dbConnInitialized
+}
+
+func (c *Cache) SetDbConnInitialized(initialized bool) {
```
jlpadilla, line 36: Set should be private, it can only be changed from inside the package.
Get is public.
```suggestion
func (c *Cache) setDbConnInitialized(initialized bool) {
```

jlpadilla (approved):
/lgtm
/hold

pkg/rbac/dbAvailability.go
```diff
@@ -0,0 +1,21 @@
...
+	"net/http"
+
+	"k8s.io/klog/v2"
+)
+
+func (c *Cache) CheckDBAvailability(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// if postgres db is not setup, return error
+		if !c.dbConnInitialized {
+			http.Error(w, "Unable to establish connection with database.", http.StatusServiceUnavailable)
```
jlpadilla, line 13: Let's log this error too.
```suggestion
			klog.Warn("Unable to handle request because we couldn't establish connection with database.") 
			http.Error(w, "Unable to establish connection with database.", http.StatusServiceUnavailable)
```

pkg/rbac/dbAvailability.go
```diff
@@ -0,0 +1,21 @@
...
+			return
+		}
+
+		klog.V(6).Info("Connection with database successful!")
```
jlpadilla, line 17: This log message could be misleading because we aren't testing the connection.
I suggest that we remove it and only log on the error path above.
```suggestion
		klog.V(6).Info("Connection with database successful!")
```

jlpadilla (approved):
/lgtm

PR #183, opened 2023-10-13
ACM-7857 Deep copy UserData to avoid concurrency problems

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-7857

### Description of changes
- Fixes map concurrency error. UserData had maps that weren't correctly copied and contained references to the objects in the cache. This caused concurrency problems if the cache gets updated while using the copy on UserData.


SherinV (approved):
/lgtm
/hold

PR #185, opened 2023-10-16
[release-2.8]  Ignore gosec G602 because of false positive.

jlpadilla:
### Related Issue
N/A

### Description of changes
- Ignore gosec G6002 which causes prow build to fail.  The main branch already has this change.


SherinV (approved):
/lgtm
/hold

PR #189, opened 2023-12-19
ACM:8774: Implement federated global search

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-8774

### Description of changes
- Initial implementation of federated global search.

### Limitations
- Does not support query aliases. https://graphql.org/learn/queries/#aliases
- HTTPS requests using `--insecure`
- LIMIT and SORT not honored in merged results.
- Relationships query not yet supported.


Makefile
```diff
@@ -72,12 +71,47 @@ test-scale-setup: ## Creates the search-api route in the current target cluster.
...
 
 test-send: ## Sends a graphQL query using cURL for development testing.
-	URL=https://localhost:4010/searchapi/graphql \
-	TOKEN=$(shell oc whoami -t) \
 	curl --insecure --location --request POST ${URL} \
-	--header "Authorization: Bearer ${TOKEN}" --header 'Content-Type: application/json' \
+	--header "Authorization: Bearer ${API_TOKEN}" --header 'Content-Type: application/json' \
 	--data-raw '{"query":"query q($$input: [SearchInput]) { search(input: $$input) { count items } }","variables":{"input":[{"keywords":[],"filters":[{"property":"kind","values":["ConfigMap"]}],"limit": 3}]}}' | jq
 
+# Target API URL for the test queries.
```
jlpadilla, line 78: Changes here are to send test queries easily. Some examples:
```
make send QUERY=schema FEDERATED=true
make send Q=search F=true
make send Q=searchComplete
```

pkg/federated/mergeResponses.go
```diff
@@ -0,0 +1,87 @@
...
+		// else {
+		// 	klog.V(9).Infof("SearchComplete - Skipping duplicate value: %s", prop)
+		// }
+	}
+
+	// TODO: How to handle LIMIT ?
+	// TODO: How to handle SORT ?
+}
+
+func (d *Data) mergeSearchResults(hubName string, results []SearchResult) {
```
SherinV, line 50: Should hubname be an array too, if results are from different clusters?

SherinV: Oh, this is from one hub - not all the hubs together. Got it.

pkg/federated/federated.go
```diff
@@ -0,0 +1,99 @@
...
+
+var getFedConfig = getFederationConfig
+var httpClientGetter = GetHttpClient
+
+func HandleFederatedRequest(w http.ResponseWriter, r *http.Request) {
+	klog.Info("Resolving federated search query.")
+	ctx := r.Context()
+	receivedBody, err := io.ReadAll(r.Body)
+	if err != nil {
+		klog.Errorf("Error reading request body: %s", err)
```
SherinV, line 29: Do we proceed after this err or return?

SherinV (approved):
/lgtm
/hold

pkg/federated/federated.go
jlpadilla: I’ll return with error because we can’t do anything without the body.

jlpadilla: done

PR #190, opened 2023-12-19
ACM-8949: fix number filter

SherinV:
### Related Issue

[https://issues.redhat.com/browse/ACM-8949](https://issues.redhat.com/browse/ACM-8949)

### Description of changes
- Added ...


pkg/resolver/searchHelper.go
```diff
@@ -62,6 +62,9 @@ func getWhereClauseExpression(prop, operator string, values []string) []exp.Expr
 		lhsExp = goqu.C(prop)
 	} else {
 		lhsExp = goqu.L(`"data"->>?`, prop)
+		if dataType == "number" {
+			lhsExp = goqu.L(`("data"->>?)?`, prop, goqu.L("::numeric"))
```
jlpadilla, line 66: Should we use `->` instead of `->>` ?

`->>` is equivalent to adding `::text` so it feels like we are adding an unnecessary step.

SherinV: Done

jlpadilla (approved):
/lgtm
/hold

PR #203, opened 2024-02-05
ACM-8775: Global search merge relationships

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-8775

### Description of changes
- Implement relationships for Global Search.

### DIFF
--- a/pkg/federated/mergeResponses.go
+++ b/pkg/federated/mergeResponses.go
@@ -4,7 +4,7 @@ package federated
 import "k8s.io/klog/v2"
 
 func (d *Data) mergeSearchSchema(schemaProps []string) {
-	klog.Info("Merge searchSchema results to federated response.")
+	klog.V(1).Info("Merge [searchSchema] results to federated response.")
 	d.writeLock.Lock()
 	defer d.writeLock.Unlock()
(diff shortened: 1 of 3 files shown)


pkg/federated/mergeResponses.go
```diff
@@ -86,24 +93,32 @@ func (d *Data) mergeMessages(msgs []string) {
...
+		return newItems
+	}
+
+	// Related results are grouped by kind.
+	for _, newItem := range newItems {
+		found := false
+		kind := newItem.Kind
+
+		// If the kind is found, merge the newItems with the mergedItems.
+		for index, mergedItem := range mergedItems {
```
SherinV, line 109: if the search term is not drilled down, like kind:Cluster, there might be a lot of mergedItems to loop through to find `kind`.

SherinV (approved):
Left one concern, but we can address it later, if an issue.
/lgtm
/hold

pkg/federated/mergeResponses.go
jlpadilla: Agree, I'll merge this to get it working and revisit the performance after.

PR #205, opened 2024-02-14
ACM-9302: Update federated search setup and cleanup scripts

jlpadilla:
### Related Issue

https://issues.redhat.com/browse/ACM-9302

### Description of changes
- Simplify usage of setup.sh and cleanup.sh scripts by including the yaml templates inline.
- Improve cleanup.sh to check if ManagedServiceAccount is being used in the cluster before disabling.
- Update README.md

### DIFF
--- a/pkg/federated/README.md
+++ b/pkg/federated/README.md
@@ -8,7 +8,11 @@ Use federated search to query and combine results from multiple Red Hat Advanced
     - Managed Hub clusters mu have RHACM 2.9.0 or later
 
 ## Setup
-Execute the script at `./setup.sh` to configure Global Search on the Global Hub cluster.  
+Execute the script at `./setup.sh` to configure Global Search on the Global Hub cluster. 
+
+```bash
+bash <(curl -s https://raw.githubusercontent.com/stolostron/search-v2-api/main/pkg/federated/setup.sh) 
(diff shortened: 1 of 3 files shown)


pkg/federated/README.md
```diff
@@ -8,7 +8,11 @@ Use federated search to query and combine results from multiple Red Hat Advanced
     - Managed Hub clusters mu have RHACM 2.9.0 or later
 
 ## Setup
-Execute the script at `./setup.sh` to configure Global Search on the Global Hub cluster.  
+Execute the script at `./setup.sh` to configure Global Search on the Global Hub cluster. 
+
```
SherinV, line 12: Should we note that this script will need to be re-executed if a new cluster if imported after the setup?

SherinV (approved):
/lgtm
/hold

pkg/federated/README.md
jlpadilla: It is on the "Notes" below.
> You must re-run this script when a Managed Hub is added.
Maybe we could add a "Limitations" section to make sure we set the correct expectations. Thoughts?

SherinV: Oh, missed the notes. That should be fine.

PR #218, opened 2024-05-14
update to go1.21

SherinV:
### Related Issue

### Description of changes
- Added ...


jlpadilla (approved):
/lgtm
/hold
We also need to update prow and downstream before we can use the new slices library.

PR #221, opened 2024-05-23
ACM 10019 - enable managedHub Search

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-10019

### Description of changes
- check if to proceed with Search based on ManagedHub filter

### DIFF
--- a/pkg/resolver/search.go
+++ b/pkg/resolver/search.go
@@ -12,11 +12,13 @@ import (
 	"github.com/driftprogramming/pgxpoolmock"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stolostron/search-v2-api/graph/model"
+	"github.com/stolostron/search-v2-api/pkg/config"
 	db "github.com/stolostron/search-v2-api/pkg/database"
 	"github.com/stolostron/search-v2-api/pkg/metrics"
 	"github.com/stolostron/search-v2-api/pkg/rbac"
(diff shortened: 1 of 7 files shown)


pkg/resolver/search.go
```diff
@@ -77,6 +98,9 @@ func (s *SearchResult) Count() (int, error) {
 func (s *SearchResult) Items() ([]map[string]interface{}, error) {
 	s.wg.Add(1)
 	defer s.wg.Done()
+	if !s.isManagedHub() { // if current hub is not part of managedHub filter, stop search
+		return nil, nil
```
jlpadilla, line 102: There's a slight difference with returning nil vs. an empty []map[string]interface{}.  I think an empty list is slightly better and safer.

pkg/resolver/search.go
```diff
@@ -65,7 +67,26 @@ func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, e
 
 }
 
+// Stop search if managedHub is in filters and current hub name is not in values.
+// Otherwise, proceed with the search.
+func (s *SearchResult) isManagedHub() bool {
```
jlpadilla, line 72: maybe a more specific name could be `matchesManagedHubFilter()`

jlpadilla (approved):
/lgtm
/hold

SherinV:
Need to merge https://github.com/stolostron/search-v2-operator/pull/191 too.

PR #226, opened 2024-06-05
ACM 11654 - 5. Partial match managed hub

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-11654

### Description of changes
- Added ...

### DIFF
--- a/pkg/resolver/search.go
+++ b/pkg/resolver/search.go
@@ -77,9 +77,14 @@ func (s *SearchResult) matchesManagedHubFilter() bool {
 
 			opValueMap := matchOperatorToProperty("string", map[string][]string{}, PointerToStringArray(filter.Values), filter.Property)
 			klog.V(5).Infof("Extract operator from managedHub filter: %+v \n", opValueMap)
+			proceedWithSearch := make([]bool, len(opValueMap))
+			count := 0
 			for key, values := range opValueMap {
-				return processOpValueMapManagedHub(key, values)
+				klog.Info("Processing ", key, " values ", values, " in opValueMap for managedHub filter")
+				proceedWithSearch[count] = processOpValueMapManagedHub(key, values)
+				count++
 			}
+			return anyTrue(proceedWithSearch) // check if any of the patterns match
 		}
 	}
 	klog.V(4).Infof("managedHub filter not present. Proceeding with Search")
--- a/pkg/resolver/search.go
+++ b/pkg/resolver/search.go
@@ -77,14 +77,14 @@ func (s *SearchResult) matchesManagedHubFilter() bool {
 
 			opValueMap := matchOperatorToProperty("string", map[string][]string{}, PointerToStringArray(filter.Values), filter.Property)
 			klog.V(5).Infof("Extract operator from managedHub filter: %+v \n", opValueMap)
-			proceedWithSearch := make([]bool, len(opValueMap))
-			count := 0
+
 			for key, values := range opValueMap {
 				klog.Info("Processing ", key, " values ", values, " in opValueMap for managedHub filter")
-				proceedWithSearch[count] = processOpValueMapManagedHub(key, values)
-				count++
+				if processOpValueMapManagedHub(key, values) {
+					return true
+				}
 			}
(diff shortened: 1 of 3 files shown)


pkg/resolver/search.go
```diff
@@ -67,19 +66,28 @@ func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, e
...
+			opValueMap := matchOperatorToProperty("string", map[string][]string{}, PointerToStringArray(filter.Values), filter.Property)
+			klog.V(5).Infof("Extract operator from managedHub filter: %+v \n", opValueMap)
+			proceedWithSearch := make([]bool, len(opValueMap))
+			count := 0
+			for key, values := range opValueMap {
+				klog.Info("Processing ", key, " values ", values, " in opValueMap for managedHub filter")
+				proceedWithSearch[count] = processOpValueMapManagedHub(key, values)
+				count++
 			}
+			return anyTrue(proceedWithSearch) // check if any of the patterns match
```
jlpadilla, line 87: Is this the same? Why not return as soon as we find a pattern match?

```suggestion
			for key, values := range opValueMap {
				klog.Info("Processing ", key, " values ", values, " in opValueMap for managedHub filter")
				if processOpValueMapManagedHub(key, values) {
				  return true
				}
			}
			return false
```

pkg/resolver/searchHelper.go
```diff
@@ -459,3 +462,86 @@ func matchOperatorToProperty(dataType string, opValueMap map[string][]string,
...
+		}
+	}
+	klog.V(4).Infof("%s not in managedHub filter %+v", config.Cfg.HubName, values)
+	return false, nil
+}
+
+// processOpValueMapManagedHub processes the key-value pair for a managedHub filter.
+// It handles different key cases such as "!", "!=", "=", "!:*", "!=:*", and "=:*".
+// It returns a boolean indicating whether the search should proceed based on the evaluation of the key and values.
+func processOpValueMapManagedHub(key string, values []string) bool {
```
jlpadilla, line 499: **Suggestion**: We can reduce the complexity moving the log to the end and including the inputs and results like this.

```
func processOpValueMapManagedHub(key string, values []string) bool {
	result := false
	switch key {
	case "!", "!=":
		result = !slices.Contains(values, config.Cfg.HubName)
	case "=":
		result = slices.Contains(values, config.Cfg.HubName)
	case "!:*", "!=:*":
		match, err := partialMatchStringPattern(values)
		if err != nil {
			klog.Errorf("Error processing partial match for ManagedHub filter:", err)
			return false
		}
		result = !match // Return the inverse of match to indicate search should not proceed if there is a partial match
	case "=:*":
		match, err := partialMatchStringPattern(values)
		if err != nil {
			klog.Errorf("Error processing partial match for ManagedHub filter:", err)
			return false
		}
		return match // Return match to indicate search should proceed if there is a partial match
	}

	klog.V(4).Infof("ManagedHub filter hubname: %s operation: %s values: %+v  result: %t",
	config.Cfg.HubName, key, values, result)

	return result
}
```

pkg/resolver/search.go
```diff
@@ -67,19 +66,29 @@ func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, e
...
-				return false
+			klog.V(5).Infof("managedHub filter: %s values: %+v \n", filter.Property,
+				PointerToStringArray(filter.Values))
+
+			opValueMap := matchOperatorToProperty("string", map[string][]string{},
+				PointerToStringArray(filter.Values), filter.Property)
+			klog.V(5).Infof("Extract operator from managedHub filter: %+v \n", opValueMap)
+
+			for key, values := range opValueMap {
+				klog.Info("Processing ", key, " values ", values, " in opValueMap for managedHub filter")
```
jlpadilla, line 83: log verbosity
```suggestion
				klog.V(5).Info("Processing ", key, " values ", values, " in opValueMap for managedHub filter")
```

jlpadilla (approved):
/lgtm
/hold
Added a minor comment for log verbosity.

PR #279, opened 2024-11-18
ACM-15250: Add GraphQL subscription for search results

zlayne:
### Related Issue

https://issues.redhat.com/browse/ACM-15250

### Description of changes
- Added ...


pkg/resolver/searchSubscription.go
```diff
@@ -0,0 +1,46 @@
...
+		// Handle deregistration of the channel here. Note the `defer`
+		defer close(ch)
+
+		for {
+			// Send the search results every 10 seconds.
+			time.Sleep(10 * time.Second)
+			fmt.Println("SearchSubscription new poll interval")
+			
+			// TODO 2nd return item is error... needs to be handled
+			searchResult, _ := Search(ctx, input)
```
smcavey, line 27: Are you saying we're getting an error here or just that we need to implement handling? Because I'm not encountering any errors in testing

zlayne: Just that I need to handle/log it.

pkg/resolver/searchSubscription.go
```diff
@@ -0,0 +1,46 @@
...
+
+			// The subscription may have got closed due to the client disconnecting.
+			// Hence we do send in a select block with a check for context cancellation.
+			// This avoids goroutine getting blocked forever or panicking,
+			select {
+			case <-ctx.Done(): // This runs when context gets cancelled. Subscription closes.
+				fmt.Println("Subscription Closed")
```
smcavey, line 34: We'll just want to convert this to a log instead of Println before merge

smcavey: As well as probably err log the above unhandled err. Otherwise from testing and what I can see in the graphql subscription docs looks good.

graph/schema.graphqls
```diff
@@ -3,28 +3,39 @@ Copyright Contributors to the Open Cluster Management project.
...
+}
+
+"""
+`Subscriptions` supported by the Search Query API.
+"""
+type Subscription {
+  """
+  searchSubscription will return a stream of `SearchResult` objects.
+  """
+  searchSubscription(input: [SearchInput]): [SearchResult]
```
jlpadilla, line 18: The name `searchSubscription` sounds redundant.  Can we name it just `search` or will it create a conflict with the Query with the same name?

graph/schema.graphqls
```diff
@@ -3,28 +3,39 @@ Copyright Contributors to the Open Cluster Management project.
...
   query: Query
+  subscription: Subscription
+}
+
+"""
+`Subscriptions` supported by the Search Query API.
+"""
+type Subscription {
+  """
+  searchSubscription will return a stream of `SearchResult` objects.
```
jlpadilla, line 16: How about
```
This subscription is experimental and must not be used by clients at this time.
Returns a stream of `SearchResult` objects.
```

graph/schema.graphqls
```diff
@@ -3,28 +3,39 @@ Copyright Contributors to the Open Cluster Management project.
...
 Search Query API.
 """
-schema { 
+schema {
   query: Query
+  subscription: Subscription
+}
+
+"""
+`Subscriptions` supported by the Search Query API.
```
jlpadilla, line 12: Avoid the word "supported". Maybe
"Subscriptions implemented by the Search Query API."

pkg/resolver/searchSubscription.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// You can (and probably should) handle your channels in a central place outside of `schema.resolvers.go`.
+	// For this example we'll simply use a Goroutine with a simple loop.
+	go func() {
+		// Handle deregistration of the channel here. Note the `defer`
+		defer close(ch)
+
+		for {
+			// Send the search results every 10 seconds.
+			time.Sleep(10 * time.Second)
```
jlpadilla, line 23: Let's make this polling interval configurable.
You can define a new env in `/pkg/config/config.go`

I'm also wondering if we should allow the client to override when creating the subscription. For this case maybe it would make sense to have a `MIN_SUBSCRIPTION_REFRESH_INTERVAL` and we allow the client to extend it. **This part can be done in the next iteration.**

pkg/resolver/searchSubscription.go
```diff
@@ -0,0 +1,46 @@
...
+			fmt.Println("SearchSubscription new poll interval")
```
jlpadilla, line 24: use `klog.Info()`

pkg/resolver/searchSubscription.go
```diff
@@ -0,0 +1,46 @@
...
+			// This avoids goroutine getting blocked forever or panicking,
+			select {
+			case <-ctx.Done(): // This runs when context gets cancelled. Subscription closes.
+				fmt.Println("Subscription Closed")
+				// Handle deregistration of the channel here. `close(ch)`
+				return // Remember to return to end the routine.
+			
+			case ch <- searchResult: // This is the actual send.
+				// Our message went through, do nothing	
+			}
```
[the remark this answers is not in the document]
jlpadilla: Is there 2-way communication? I'm wondering if the client could send anything to indicate "I'm still listening, keep the subscription open"
If that's possible, then we could implement it as `MAX_SUBSCRIPTION_IDLE_TIME`.

graph/schema.graphqls
zlayne: Need to test a little more, but this seems to be fine as search is contained within the Subscription type.

zlayne:
@jlpadilla @smcavey Let me know if the new changes look good. I am working on unit tests now. 
Also the lint issue is in the generated.go file. Do we ignore those?

smcavey:
Can't find any open issues with gqlgen regarding the gosec issues with the handled errors we're seeing here. I'd rather not see these issues on each and every pr. I thought by default gosec was supposed to ignore generated files, but it seems its scanning them anyway.

```
diff --git a/Makefile b/Makefile
index 9f48eb7..231147f 100644
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ lint: ## Run lint and gosec tools.
        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "${GOPATH}/bin" v1.61.0
        CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=3m
        go mod tidy
-       gosec ./...
+       gosec -exclude-generated ./...
```

Or getting more specific with `-exclude-dir` to just skip this one can get the lint job passing.

pkg/resolver/searchSubscription.go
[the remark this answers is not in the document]
smcavey: It does seem that the subscription connection will close the request context when the client is closed: https://stackoverflow.com/questions/77500863/how-client-sends-close-signal-to-graphql-subscription-server

There also appears to be a keepalivepinginterval attribute on the websocket: https://github.com/99designs/gqlgen/blob/master/graphql/handler/transport/websocket.go#L29 that sends pings from server to client and can detect lost or disconnected sessions: https://github.com/spring-projects/spring-graphql/issues/534 which could help ensure we're not doing more work when the client has disconnected. But only `MAX_SUBSCRIPTION_IDLE_TIME` would stop the server work if the user starts a subscription and then walks away for a while with the client open.

graph/schema.resolvers.go
```diff
@@ -41,7 +41,17 @@ func (r *queryResolver) Messages(ctx context.Context) ([]*model.Message, error)
 	return resolver.Messages(ctx)
 }
 
+// SearchSubscription is the resolver for the search field.
+func (r *subscriptionResolver) Search(ctx context.Context, input []*model.SearchInput) (<-chan []*resolver.SearchResult, error) {
+	klog.Infoln("Received search query subscription")
```
jlpadilla, line 46: Increase verbosity to avoid flooding the log.
```suggestion
	klog.V(3).Info("Received search query subscription")
```

pkg/resolver/searchSubscription.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// You can (and probably should) handle your channels in a central place outside of `schema.resolvers.go`.
+	// For this example we'll simply use a Goroutine with a simple loop.
+	go func() {
+		timeout := time.After(time.Duration(config.Cfg.SubscriptionPollTimeout) * time.Minute)
+		// Handle deregistration of the channel here. Note the `defer`
+		defer close(ch)
+
+		for {
+			klog.Infof("Search subscription new poll interval")
```
jlpadilla, line 25: 1. Change to `klog.Info()`.   FYI, Infof() is used when including variables in the log.
2. Increase verbosity with `V(3)` or higher.

graph/generated/generated.go
```diff
@@ -252,28 +290,40 @@ Copyright Contributors to the Open Cluster Management project.
 
 Search Query API.
 """
-schema {
```
jlpadilla, line 255: Please regenerate to fix format.

pkg/resolver/searchSubscription.go
```diff
@@ -0,0 +1,55 @@
...
+			case <-ctx.Done(): // This runs when context gets cancelled. Subscription closes.
+				klog.V(3).Info("Search subscription Closed")
+				return // Remember to return to end the routine.
+
+			case <-timeout: // This runs when timeoout is hit. Subscription closes.
+				klog.V(3).Info("Subscription timeout reached. Closing connection.")
+				return // Remember to return to end the routine.
+			
+			case ch <- searchResult: // This is the actual send.
+				// Our message went through, do nothing
```
jlpadilla, line 45: I want to take a close look at what is actually being sent. We don't want to send the entire searchResult object every 10 seconds because it will be a lot of unnecessary data. We just want to send the delta.
I hope the library implementation takes care of it, but I'd like to confirm it.

zlayne: With the current logic we would be sending the entire search result back. I will need to look into getting just the delta, and how that would need to be handled in frontend as well.

jlpadilla: Let's talk about this one. This is good progress, but I'm thinking we must resolve this soon to avoid having to re-do the front-end changes.

jlpadilla: Ok, from the function I see we can only add more []*SearchResult to the channel, so this would be the entire object.  This will also cause issues if there's multiple SearchInput in the subscription. All of the inputs will need to get evaluated.

```
func SearchSubscription(ctx context.Context, input []*model.SearchInput) (<-chan []*SearchResult, error) {
	ch := make(chan []*SearchResult)
```

Let's continue learning. :-)

pkg/config/config.go
```diff
@@ -17,36 +17,39 @@ var Cfg = new()
...
+	Features                 featureFlags     // Enable or disable features.
+	Federation               federationConfig // Federated search configuration.
+	HttpPort                 int
+	PlaygroundMode           bool   // Enable the GraphQL Playground client.
+	PodNamespace             string // Kubernetes namespace where the pod is running.
+	QueryLimit               uint   // The default LIMIT to use on queries. Client can override.
+	RelationLevel            int    // The number of levels/hops for finding relationships for a particular resource
+	SlowLog                  int    // Logs when queries are slower than the specified time duration in ms. Default 300ms
+	SubscriptionPollInterval int    // Number of seconds between subscription polls
+	SubscriptionPollTimeout  int    // Minutes a subscription will stay open before timeout
```
jlpadilla, line 46: Let's try to keep the time units consistent.
Other time units are in ms, at least these 2 should be the same.

pkg/config/config.go
```diff
@@ -90,6 +93,7 @@ func new() *Config {
 		DevelopmentMode:     DEVELOPMENT_MODE,
 		Features: featureFlags{
 			FederatedSearch: getEnvAsBool("FEATURE_FEDERATED_SEARCH", false), // In Dev mode default to true.
+			SubscriptionEnabled: getEnvAsBool("SUBSCRIPTION_ENABLED", false),
```
jlpadilla, line 96: Maybe rename to `FEATURE_SUBSCRIPTION` for consistency with the above.

zlayne:
@jlpadilla using the name `experimentalSearch` for the subscription. Let me know if a different name works better.

pkg/config/config.go
jlpadilla: Maybe call it `SUBSCRIPTION_REFRESH_INTERVAL` so it still works when we change the implementation.

jlpadilla:
/lgtm
/override ci/prow/sonar-pre-submit

This PR doesn't meet the test coverage because we expect significant changes to the implementation in the next iteration.

jlpadilla (approved):
/lgtm

PR #281, opened 2024-11-19
Disable pgcpool statement cache to reduce memory usage

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-13201

### Description of changes
- Disables the pgxpool statement cache. This disables the use of prepared statements which consume significant amount of memory in postgres.
- More information: https://aws.amazon.com/blogs/database/resources-consumed-by-idle-postgresql-connections/


smcavey:
Observing a bit more stability and slightly less memory utilization by search-postgres pod when applying these changes over https://github.com/stolostron/search-v2-api/pull/280: 
Screenshot from 2024-11-20 12-16-03
*pictures memory stability after 12:00 PM when changes applied to search-api deployment*

smcavey:
Leaving https://github.com/jackc/pgx/discussions/1535 here for if/when we upgrade to pgx v5 and BuildStatementCache get replaced by QueryExecMode and we need to change.

smcavey:
Memory utilization looking low and stable still. Looks good to me.

smcavey (approved):

PR #395, opened 2025-03-29
2.12 fix cves

smcavey:
fix cves from https://github.com/stolostron/search-v2-api/pull/384


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/stolostron/search-v2-api
 
-go 1.22
+go 1.23.0
+
+toolchain go1.23.7
```
jlpadilla, line 5: Do we need this line?

smcavey: Shouldn't. Was trying to get a head start on the next PR but think I need to build fresh after rebasing on https://github.com/stolostron/search-v2-api/pull/393.

PR #459, opened 2025-04-22
fix sast-coverity-check-oci-ta by adding missing image-digest

smcavey:
### Description of changes
- Added image-digest param to sast-coverity-check-oci-ta to fix https://github.com/stolostron/search-v2-api/pull/444


jlpadilla (approved):
/lgtm

PR #499, opened 2025-05-12
ACM-20492 remove hard-coded local-cluster

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-20492

### Description of changes
- Removed hard-coded "local-cluster" by grabbing cluster name from database and by grabbing all managed cluster names that don't have local-cluster label

### DIFF
--- a/pkg/resolver/rbacHelper.go
+++ b/pkg/resolver/rbacHelper.go
@@ -190,6 +190,8 @@ func matchManagedCluster(managedClusters []string) exp.BooleanExpression {
 
 	if len(managedClusters) == 1 && managedClusters[0] == "*" {
 		klog.V(2).Infof("user has access to all managed clusters")
+		// SQL with subquery tested to be ~33% faster compared to returning goqu.L("(data ? '_hubClusterResource')").Eq(false)
+		// so, intentionally going with larger query
 		return goqu.C("cluster").Neq(subQuery)
 	}
 	//managed clusters
--- a/pkg/resolver/rbacHelper.go
+++ b/pkg/resolver/rbacHelper.go
@@ -184,9 +184,13 @@ func matchHubCluster(userrbac rbac.UserData, userInfo v1.UserInfo) exp.Expressio
 //	( cluster IN ['a', 'b', ...] )
 
 func matchManagedCluster(managedClusters []string) exp.BooleanExpression {
+	subQuery := goqu.From("search.resources").
+		Select("cluster").
+		Where(goqu.L("(data ? '_hubClusterResource')").Eq(true)).Limit(1)
+
 	if len(managedClusters) == 1 && managedClusters[0] == "*" {
 		klog.V(2).Infof("user has access to all managed clusters")
-		return goqu.C("cluster").Neq("local-cluster")
+		return goqu.C("cluster").Neq(subQuery)
 	}
 	//managed clusters
 	return goqu.C("cluster").Eq(goqu.Any(pq.Array(managedClusters)))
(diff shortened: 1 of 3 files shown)


pkg/resolver/rbacHelper.go
```diff
@@ -184,9 +184,13 @@ func matchHubCluster(userrbac rbac.UserData, userInfo v1.UserInfo) exp.Expressio
...
 
 func matchManagedCluster(managedClusters []string) exp.BooleanExpression {
+	subQuery := goqu.From("search.resources").
+		Select("cluster").
+		Where(goqu.L("(data ? '_hubClusterResource')").Eq(true)).Limit(1)
+
 	if len(managedClusters) == 1 && managedClusters[0] == "*" {
 		klog.V(2).Infof("user has access to all managed clusters")
-		return goqu.C("cluster").Neq("local-cluster")
+		return goqu.C("cluster").Neq(subQuery)
```
jlpadilla, line 193: This works, but I was expecting this to resolve to 
```go
return goqu.L("(data ? '_hubClusterResource')").Eq(false)
```

I'd like to check what's the difference in performance of the 2 options.

smcavey: `search=> explain analyze select uid from "search"."resources" where (("data" ->> 'kind' ilike any ('{"template"}')) and ((data ? '_hubClusterResource') is false)) limit 1000;
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..25706.53 rows=1 width=50) (actual time=14.426..14.427 rows=0 loops=1)
   ->  Seq Scan on resources  (cost=0.00..25706.53 rows=1 width=50) (actual time=14.426..14.426 rows=0 loops=1)
         Filter: (((data ? '_hubClusterResource'::text) IS FALSE) AND ((data ->> 'kind'::text) ~~* ANY ('{template}'::text[])))
         Rows Removed by Filter: 5939
 Planning Time: 0.268 ms
 Execution Time: 14.438 ms
(6 rows)

search=> explain analyze select uid from "search"."resources" where (("data" ->> 'kind' ilike any ('{"template"}')) and ("cluster" not in (select "cluster" from "search"."resources" where ((data ? '_hubClusterResource') is true) limit 1))) limit 1000;
                                                                QUERY PLAN                                                                 
-------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=4.33..25710.85 rows=1 width=50) (actual time=8.921..8.922 rows=0 loops=1)
   ->  Seq Scan on resources  (cost=4.33..25710.85 rows=1 width=50) (actual time=8.920..8.921 rows=0 loops=1)
         Filter: ((NOT (hashed SubPlan 1)) AND ((data ->> 'kind'::text) ~~* ANY ('{template}'::text[])))
         Rows Removed by Filter: 5940
         SubPlan 1
           ->  Limit  (cost=0.00..4.32 rows=1 width=13) (actual time=0.003..0.003 rows=1 loops=1)
                 ->  Seq Scan on resources resources_1  (cost=0.00..25684.25 rows=5939 width=13) (actual time=0.002..0.003 rows=1 loops=1)
                       Filter: ((data ? '_hubClusterResource'::text) IS TRUE)
 Planning Time: 0.207 ms
 Execution Time: 9.007 ms
(10 rows)`

smcavey: Estimated costs are basically the same. The first reads simpler, the second in actuality was remotely faster, I'm guessing because we have to go into the jsonb best case once

jlpadilla (approved):
/lgtm

PR #520, opened 2025-05-30
ACM-18446 missing FROM jsonb_each_text

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-18446

### Description of changes
- Added missing FROM jsonb_each_text to searchComplete and searchSchema when input has keywords. The resulting sql.
SQL before: `SELECT DISTINCT "data"->'label' FROM "search"."resources" WHERE (("value" ILIKE '%search%') AND "data"->'kind'?('Pod') AND ("data"->'label' IS NOT NULL)) ORDER BY "data"->'label' ASC LIMIT 10000` <- "value" doesn't exist because no jsonb_each_text
SQL after: `SELECT DISTINCT "data"->'label' FROM "search"."resources", jsonb_each_text("data") WHERE (("value" ILIKE '%search%') AND "data"->'kind'?('Pod') AND ("data"->'label' IS NOT NULL)) ORDER BY "data"->'label' ASC LIMIT 10000` <- "value" exists now :)

### DIFF
--- a/pkg/resolver/search.go
+++ b/pkg/resolver/search.go
@@ -347,12 +347,12 @@ func WhereClauseFilter(ctx context.Context, input *model.SearchInput,
 	var err error
 
 	if len(input.Keywords) > 0 {
-		// Sample query: SELECT COUNT("uid") FROM "search"."resources", jsonb_each_text("data")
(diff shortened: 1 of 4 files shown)


zlayne:
I believe this change will also fix the following bug: https://issues.redhat.com/browse/ACM-21036

PR #572, opened 2025-07-09
remove konflux 2.14 config from 2.15

smcavey:
### Description of changes
- remove 2.14 konflux config files from 2.15 so we don't break anything


jlpadilla (approved):
/lgtm

PR #605, opened 2025-08-11
[main] hermetic, multi-arch, params konflux fix

smcavey:
### Description of changes
- Added multi-arch builds back, hermetic builds, other params konflux missed with the 2.15 config


jlpadilla (approved):
/lgtm

PR #641, opened 2025-09-29
expand readiness check to check rbac cache and db

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-21444

### Description of changes
- Added expanded readiness check to do a basic rbac cache and database connectivity check.

### DIFF
--- a/pkg/rbac/cache.go
+++ b/pkg/rbac/cache.go
@@ -37,27 +37,6 @@ func (c *Cache) SetDbConnInitialized(initialized bool) {
 	c.dbConnInitialized = initialized
 }
 
-// NewMockCacheForTesting creates a new Cache instance with minimal initialization for testing.
-// This is exported specifically for use in tests that need a functional cache mock.
-func NewMockCacheForTesting(healthy bool, pool pgxpoolmock.PgxPool) *Cache {
-	cache := &Cache{
-		tokenReviews:     map[string]*tokenReviewCache{},
-		tokenReviewsLock: sync.Mutex{},
-		users:            map[string]*UserDataCache{},
-		usersLock:        sync.Mutex{},
-		shared: SharedData{
-			disabledClusters: map[string]struct{}{},
-			managedClusters:  map[string]struct{}{"test-cluster": {}}, // Add at least one cluster
-			namespaces:       []string{"default"},                     // Add at least one namespace
-			pool:             pool,
-		},
-		pool:              pool,
-		dbConnInitialized: healthy,
-	}
-
-	return cache
-}
-
 // Initialize the cache as a singleton instance.
 var cacheInst = Cache{
 	tokenReviews:     map[string]*tokenReviewCache{},


pkg/server/healthProbes.go
```diff
@@ -15,8 +19,32 @@ func livenessProbe(w http.ResponseWriter, r *http.Request) {
...
 }
 
-// ReadinessProbe checks if database is available.
+// ReadinessProbe checks if rbac cache and database are available.
 func readinessProbe(w http.ResponseWriter, r *http.Request) {
 	klog.V(5).Info("readinessProbe")
+	ctx, cancel := context.WithTimeout(r.Context(), 5*time.Second)
+	defer cancel()
+
+	pool := database.GetConnPool(ctx)
```
jlpadilla, line 28: This context is used to create the connection, but it is cancelled shortly after. I wonder if this could cause problems.

```
conn, err := pgxpool.ConnectConfig(ctx, config)
```

smcavey: Reading the pgx code and implementation, GetConnPool the timeout doesn't get respected. We have to use a ConnectTimeout setting on the connection Config. This is available with newer versions. Ping does, though. Using the context WithTimeout here won't actually timeout any connections.

pkg/server/healthProbes.go
```diff
@@ -15,8 +19,52 @@ func livenessProbe(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		if !cache.IsHealthy() {
+			cacheDone <- fmt.Errorf("RBAC cache unavailable")
+			return
+		}
+		cacheDone <- nil
+	}()
+
+	select {
```
jlpadilla, line 53: This select will wait for 1 of the 3 conditions.
We need to wait for both `dbDone` AND `cacheDone`.

jlpadilla: 🤖 AI suggestion

```go
var dbErr, cacheErr error

  // Wait for both checks
  for i := 0; i < 2; i++ {
      select {
      case err := <-dbDone:
          dbErr = err
      case err := <-cacheDone:
          cacheErr = err
      case <-ctx.Done():
          http.Error(w, "Readiness probe timed out", http.StatusServiceUnavailable)
          return
      }
  }

  // Check results
  if dbErr != nil {
      http.Error(w, "Database: "+dbErr.Error(), http.StatusServiceUnavailable)
      return
  }
  if cacheErr != nil {
      http.Error(w, "RBAC cache: "+cacheErr.Error(), http.StatusServiceUnavailable)
      return
  }

  fmt.Fprint(w, "OK")
```

smcavey: Just pushed a change, waiting for both to complete, or timeout to happen first. If both complete without timeout, we potentially concat their errs to a single err to report if one or both checks failed.

pkg/rbac/cache.go
```diff
@@ -70,3 +70,28 @@ func GetCache() *Cache {
...
+	}
+
+	// Check if we have a database pool
+	if c.pool == nil {
+		klog.V(3).Info("RBAC cache unhealthy: database pool is nil")
+		return false
+	}
+
+	// Check if shared data has been populated
+	hasData := c.shared.managedClusters != nil && c.shared.namespaces != nil
```
jlpadilla, line 89: **🤖 From Claude**

Race Condition Potential
  - Reads c.shared without locking
  - c.shared is likely modified by StartBackgroundValidation concurrently
  - Fix: Add proper mutex locking in IsHealthy() or document why it's safe

pkg/server/healthProbes_test.go
```diff
@@ -3,9 +3,11 @@
 package server
```
jlpadilla, line 3: **🤖 From Claude**

Recommendation: Add comprehensive test cases:
```go
  func TestReadinessProbe_Success(t *testing.T) { /* mock healthy state */ }
  func TestReadinessProbe_DBDown(t *testing.T) { /* mock DB failure */ }
  func TestReadinessProbe_CacheUnhealthy(t *testing.T) { /* mock cache not ready */ }
  func TestReadinessProbe_Timeout(t *testing.T) { /* mock slow response */ }
```

pkg/rbac/cache.go
```diff
@@ -33,10 +33,31 @@ func (c *Cache) GetDbConnInitialized() bool {
...
 }
 
-func (c *Cache) setDbConnInitialized(initialized bool) {
+func (c *Cache) SetDbConnInitialized(initialized bool) {
 	c.dbConnInitialized = initialized
 }
 
+// NewMockCacheForTesting creates a new Cache instance with minimal initialization for testing.
+// This is exported specifically for use in tests that need a functional cache mock.
+func NewMockCacheForTesting(healthy bool, pool pgxpoolmock.PgxPool) *Cache {
```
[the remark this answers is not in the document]
smcavey: I won't be able to import that from the _test file in another package. But I can create a test utility file in the rbac package to import it from there and keep it out of this file if desirable.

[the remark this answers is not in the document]
smcavey: I was able to simplify. The mock cache constructor func is in healthProbes_test. i was able to use the real rbac.GetCache() func and directly set db conn initialized like done in cache_test.go

PR #642, opened 2025-10-03
ACM-21444 prevent unbound connection growth

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-21444

### Description of changes
- Added timeout on request context. Considerations to be made on timing and how the chosen timeouts work with other config timings, e.g. DBMaxConnIdleTime, DBMaxConnLifeTime, etc.

### DIFF
--- a/pkg/server/timeoutMiddleware.go
+++ b/pkg/server/timeoutMiddleware.go
@@ -10,7 +10,7 @@ import (
 func TimeoutHandler(timeout time.Duration) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			klog.V(5).Infof("Applying %v timeout for HTTP request: %s %s", timeout, r.Method, r.URL.Path)
+			klog.V(8).Infof("Applying %v timeout for HTTP request: %s %s", timeout, r.Method, r.URL.Path)
 			handler := http.TimeoutHandler(next, timeout, "Request timed out")
 			handler.ServeHTTP(w, r)
 		})


pkg/config/config.go
```diff
@@ -41,9 +41,11 @@ type Config struct {
 	PodNamespace                string // Kubernetes namespace where the pod is running.
 	QueryLimit                  uint   // The default LIMIT to use on queries. Client can override. Default: 1000
 	RelationLevel               int    // The number of levels/hops for finding relationships for a particular resource
-	SlowLog                     int    // Logs queries slower than the specified duration in ms. Default: 300ms
-	SubscriptionRefreshInterval int    // Duration in seconds between subscription polls.        Default: 10 seconds
-	SubscriptionRefreshTimeout  int    // Minutes a subscription will stay open before timeout.  Default: 5 minutes
+	SlowLog                     int    // Logs queries slower than the specified duration in ms.      Default: 300ms
+	SubscriptionRefreshInterval int    // Duration in seconds between subscription polls.             Default: 10 seconds
+	SubscriptionRefreshTimeout  int    // Minutes a subscription will stay open before timeout.       Default: 5 minutes
+	RequestTimeout              int    // Seconds a request will process before timing out.           Default: 30 seconds
```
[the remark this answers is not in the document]
smcavey: Updated to 2 mins

pkg/resolver/search.go
```diff
@@ -34,10 +35,29 @@ type SearchResult struct {
...
+}
+
 func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, error) {
 	defer metrics.SlowLog("SearchResolver", 0)()
 	// For each input, create a SearchResult resolver.
 	srchResult := make([]*SearchResult, len(input))
+
+	select {
+	case <-ctx.Done():
+		return srchResult, errors.New("context exceeded deadline")
```
jlpadilla, line 57: This error message is sent in the response to the client. Let's make it more readable.
Suggestion: "Timeout processing the request."

pkg/server/contextTimeout.go
```diff
@@ -0,0 +1,31 @@
...
+			var ctx context.Context
+			var cancel context.CancelFunc
+			if websocket.IsWebSocketUpgrade(r) {
+				ctx, cancel = context.WithTimeout(r.Context(), time.Duration(cfg.StreamRequestTimeout)*time.Millisecond)
+				klog.V(5).Infof("Timeout of %v applied to stream request context", time.Duration(cfg.StreamRequestTimeout)*time.Millisecond)
+			} else {
+				ctx, cancel = context.WithTimeout(r.Context(), time.Duration(cfg.RequestTimeout)*time.Millisecond)
+				klog.V(5).Infof("Timeout of %v applied to request context", time.Duration(cfg.RequestTimeout)*time.Millisecond)
+			}
+			defer cancel()
```
jlpadilla, line 26: Maybe this is simpler?
```suggestion
			timeout := time.Duration(cfg.RequestTimeout)*time.Millisecond)
			requestType := "request"
			if websocket.IsWebSocketUpgrade(r) {
				timeout = time.Duration(cfg.StreamRequestTimeout)*time.Millisecond)
				requestType = "stream"
			} 
			
			ctx, cancel = context.WithTimeout(r.Context(), timeout)
			klog.V(5).Infof("Timeout of %v applied to %s context", timeout, requestType)
			defer cancel()
```

smcavey: That works :+1:  Applied.

pkg/rbac/userData.go
```diff
@@ -260,6 +260,14 @@ func (user *UserDataCache) getClusterScopedResources(ctx context.Context, cache
 	wg := sync.WaitGroup{}
 	lock := sync.Mutex{}
 	for res := range clusterScopedResources {
+		select {
+		case <-ctx.Done():
+			klog.Warning("Context canceled. stopping remaining SSAR checks. Examine frequency of requests and request timeout config.")
+			wg.Wait()
```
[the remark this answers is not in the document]
jlpadilla: You need to move the select block.  The code should looks like this sample.

```go
    var wg sync.WaitGroup
	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel()

	for i := 0; i < 3; i++ {
		wg.Add(1)
		go func(id int) {
			defer wg.Done()
			worker(ctx, id)
		}(i)
	}

	done := make(chan struct{})
	go func() {
		wg.Wait()
		close(done)
	}()

	select {
	case <-done:
		fmt.Println("All goroutines managed by WaitGroup completed.")
	case <-ctx.Done():
		fmt.Println("Main routine: Timeout or context cancelled:", ctx.Err())
	}
```

pkg/resolver/search.go
```diff
@@ -34,10 +35,34 @@ type SearchResult struct {
...
 func Search(ctx context.Context, input []*model.SearchInput) ([]*SearchResult, error) {
 	defer metrics.SlowLog("SearchResolver", 0)()
 	// For each input, create a SearchResult resolver.
 	srchResult := make([]*SearchResult, len(input))
+
+	select {
+	case <-ctx.Done():
+		return srchResult, errors.New("Timeout processing the request.")
+	default:
+	}
```
jlpadilla, line 64: I don't think this block works as you expect.
It only works if the timeout happens BEFORE you reach the SELECT. If there isn't a timeout, it resolves to `default:` and continues executing the code below. It won't process any timeouts that happen AFTER the SELECT is resolved.

PR #710, opened 2026-04-17
Fix shared cache deadlock causing API non-responsiveness

jlpadilla:
## Summary
- **Root cause**: Shared cache functions (`getNamespaces`, `getManagedClusters`, `getClusterScopedResources`) held write locks while making external API calls (K8s API, database queries). When these calls were slow, all concurrent requests blocked on the lock, causing request pile-up and API unresponsiveness.
- **Fix**: Release locks before external calls using RLock fast path + double-checked locking pattern. Make `PopulateSharedCache` respect context cancellation to prevent zombie goroutines from `http.TimeoutHandler`. Change `isClusterScoped` to use RLock since it only reads.
- **Test update**: Updated `Test_getResouces_usingCache` to validate that valid caches are preserved (not unnecessarily refreshed), matching the new behavior.

### Detailed Changes

**`PopulateSharedCache`**: Uses `channel + select` pattern instead of bare `wg.Wait()` so the function respects context cancellation and returns early when `http.TimeoutHandler` fires.

**`getNamespaces`, `getManagedClusters`, `getClusterScopedResources`**: All three now follow the same pattern:
1. RLock fast path — if cache is valid, return immediately
2. Release RLock, make external call without any lock held
3. Acquire write Lock, double-check cache validity (another goroutine may have refreshed it)
4. Update cache and release Lock

**`isClusterScoped`**: Changed from `Lock`/`Unlock` to `RLock`/`RUnlock` since this function only reads the map.

Resolves: ACM-28665

## Test plan
- [x] `make test` passes
- [x] `make lint` passes (golangci-lint, gosec)
- [ ] Deploy to cluster experiencing the deadlock and verify API responsiveness under load
- [ ] Monitor `getSharedNamespaces` slow log entries to confirm lock contention is resolved

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Improved cache refresh concurrency: background refreshes launched unconditionally; context cancellation returns early while background refreshes complete.
  * Reduced lock contention and avoided holding locks during external I/O; switched several checks to read-only where safe.
  * Centralized notification handling to streamline parsing, backfill, logging, and event forwarding.

* **Bug Fixes**
  * Keep subscription watches running when event buffer is full by dropping that event (and logging) instead of terminating the watch.

* **Tests**
  * Updated cache-preservation test to use pre-populated, non-expired caches.
  * Updated notification test to align with parameterized query behavior.

### DIFF
--- a/pkg/database/listener.go
+++ b/pkg/database/listener.go
@@ -269,75 +269,63 @@ func (l *Listener) listen() {
 			}
 
 			if notification != nil {
-				l.forwardNotification(notification)
+				l.mu.RLock()
+				// Ensure RUnlock is called on all paths (error and success)
+				defer l.mu.RUnlock()
+
+				klog.V(3).Infof("Received postgres event, forwarding to %d subscriptions.",
+					len(l.subscriptions))
+
+				var notificationPayload model.Event
+				err := json.Unmarshal([]byte(notification.Payload), &notificationPayload)
+				if err != nil {
+					klog.Errorf("Failed to unmarshal notification payload: %v", err)
+					continue
+				}
+
+				// If the notification payload was too large, data was truncated.
+				// We need to query the database to rebuild the data.
+				if notificationPayload.NewData == nil &&
+					(notificationPayload.Operation == "INSERT" || notificationPayload.Operation == "UPDATE") {
+					klog.V(2).Infof("Notification payload missing newData, querying the database. UID: %s", notificationPayload.UID)
+					rows, err := l.conn.Query(l.ctx, fmt.Sprintf("SELECT data FROM search.resources WHERE uid = '%s'", notificationPayload.UID))
+					if err != nil {
+						klog.Errorf("Failed to execute query: %v", err)
+						continue
+					}
+
+					// Explicitly close rows after processing to avoid resource leak
+					for rows.Next() {
+						var data map[string]any
+						err := rows.Scan(&data)
+						if err != nil {
+							klog.Errorf("Failed to scan result: %v", err)
+							continue
+						}
+						notificationPayload.NewData = data
+					}
+					rows.Close()
+
+					// Check for errors from iteration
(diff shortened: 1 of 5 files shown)


jlpadilla:
🤖 Addressed CodeRabbit feedback: added `csrCache.updatedAt = time.Now()` on both error paths in `getClusterScopedResources` so the 10-second error TTL backoff works correctly.

jlpadilla:
🧔‍♂️ 
Changes in this PR are causing the search-2e2-tests to fail.

The subscription tests show several failures when running against this PR image. Non-subscription tests pass, so the problem seems to be specific to the subscription path.

```
Test Suites: 5 failed, 9 passed, 14 total
Tests:       23 failed, 8 skipped, 37 todo, 318 passed, 386 total
Snapshots:   0 total
Time:        555.193 s
Ran all test suites.
```

jlpadilla:
🤖 **I'm an AI agent.** 

Found and fixed the root cause of the subscription test failures.

**Root cause**: In `pkg/database/listener.go`, the `listen()` function used `defer l.mu.RUnlock()` inside a `for` loop. In Go, `defer` runs when the enclosing *function* returns — not when the loop iteration ends. So after the first Postgres notification, the read lock was held permanently for the lifetime of the `listen()` goroutine.

This caused `RegisterSubscription()` and `UnregisterSubscription()` to deadlock after the first event, because both need a write lock (`mu.Lock()`). In practice: the second test's subscription was never registered, so no events were ever delivered — causing all subsequent tests to time out.

**Fix**: Extracted notification handling into `forwardNotification()`. With `defer l.mu.RUnlock()` inside a proper function (not a loop), the read lock is correctly released after each notification.

This was introduced in PR #707 and affects all subscription tests (including basic equality tests, not just operator tests).

PR #711, opened 2026-04-21
Fixes subscription e2e tests failing and problem with multiple ws connections

jlpadilla:
### Related Issue
- E2E tests failing.
- Problems with multiple subscriptions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscriptions are more resilient: if delivering an event to a client would block, the event is skipped and processing continues instead of terminating the subscription.
  * Skips cancelled subscribers without impacting others.

* **Refactor**
  * Notification handling and delivery were reorganized for non-blocking dispatch, safer backfill handling, and improved stability and observability.
  * Backfill queries now use parameterized execution.

* **Tests**
  * Added tests for notification parsing, backfill outcomes, channel-full behavior, and multi-subscriber scenarios.

### DIFF
--- a/pkg/database/listener.go
+++ b/pkg/database/listener.go
@@ -332,9 +332,16 @@ func (l *Listener) forwardNotification(notification *pgconn.Notification) {
 
 	klog.V(3).Infof("Received postgres event, forwarding to %d subscriptions.", len(l.subscriptions))
 	for _, sub := range l.subscriptions {
+		// Check Done first (without competing with the send case) to guarantee
+		// cancelled subscriptions are skipped even when the channel has space.
 		select {
 		case <-sub.Context.Done():
 			klog.V(3).Infof("Subscription %s context is done, skipping", sub.ID)
+			continue
+		default:
+		}
+		// Non-blocking send: drop the event if the channel buffer is full.
+		select {
 		case sub.Channel <- &notificationPayload:
 		default:
 			klog.Warningf("Subscription %s channel buffer is full, dropping event.", sub.ID)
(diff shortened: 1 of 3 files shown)


jlpadilla:
🤖 I'm an AI agent. Addressed the CodeRabbit comment on backfill error handling.

**Changes**: The backfill path now drops the event (instead of forwarding a partial notification with nil `newData`) on any failure:
- **Scan error** → log + return
- **`rows.Err()`** → log + return  
- **Empty result** (resource deleted before backfill) → log + return
- Used `defer rows.Close()` for resource safety
- Replaced `for+break` with `if` (one row needed; fixes staticcheck SA4004)

Added two new tests: `TestForwardNotification_BackfillNoRows` and `TestForwardNotification_BackfillScanError`.

jlpadilla:
@smcavey this PR is good for review and merge.

**search-e2e-test** pass with the PR image.

pkg/database/listener.go
```diff
@@ -269,63 +269,86 @@ func (l *Listener) listen() {
...
-						sub.Channel <- &notificationPayload
-					}
-				}
+	// If the notification payload was too large, data was truncated.
+	// We need to query the database to rebuild the data.
+	// This DB call happens without holding l.mu to avoid blocking subscription registration.
+	if notificationPayload.NewData == nil &&
+		(notificationPayload.Operation == "INSERT" || notificationPayload.Operation == "UPDATE") {
+		klog.V(2).Infof("Notification payload missing newData, querying the database. UID: %s", notificationPayload.UID)
+		rows, err := l.conn.Query(l.ctx, "SELECT data FROM search.resources WHERE uid = $1", notificationPayload.UID)
```
smcavey, line 299: Is this what the function header is calling the backfill query?

smcavey: inferred from tests that it is

